PR #1, opened 2015-11-18
No contacts in component list command result.

xychu:
JIRA: PDC-1166

### DIFF
--- a/pdc_client/plugins/component.py
+++ b/pdc_client/plugins/component.py
@@ -28,7 +28,7 @@ class GlobalComponentPlugin(PDCClientPlugin):
         self.set_command('global-component')
 
         list_parser = self.add_action('list', help='list all global components')
-        filters = ('dist_git_path contact_role email label name upstream_homepage upstream_scm_type '
+        filters = ('dist_git_path label name upstream_homepage upstream_scm_type '
                    'upstream_scm_url'.split())
         for arg in filters:
             list_parser.add_argument('--' + arg.replace('_', '-'), dest='filter_' + arg)
@@ -62,6 +62,7 @@ class GlobalComponentPlugin(PDCClientPlugin):
         filters = extract_arguments(args, prefix='filter_')
         if not filters:
             self.subparsers.choices.get('list').error('At least some filter must be used.')
+
         global_components = get_paged(self.client['global-components']._, **filters)
 
         if args.json:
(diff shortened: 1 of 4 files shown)


tests/component/data/global_component/list.json
```diff
@@ -1,38 +1 @@
...
-                "contact_role": "test_role_b",
-                "contact": {
-                    "username": "Test User",
-                    "email": "test_user@example.com"
-                }
-            }
-        ]
-    }
-]
+[{"name": "Test Global Component", "contacts": [], "labels": [{"url": "http://example.com/lable/1", "name": "test label", "description": "test label description"}], "dist_git_web_url": "http://pkgs.example.com/test_global_component", "upstream": {"homepage": "http://test_global_component.org", "scm_url": "http://test_global_component.org/git", "scm_type": "git"}, "dist_git_path": null, "id": "1"}]
```
simozhan: it is better to format it

simozhan:
+1. Looks good to me

lubomir:
For a list operation, the contacts are only used in the JSON output, right? If so, they should be only downloaded if required. However, maybe it would make sense not to include them at all.

pdc_client/plugins/component.py
```diff
@@ -64,8 +64,15 @@ def list_global_components(self, args):
...
+            component_contacts = get_paged(self.client['global-component-contacts']._,
+                                           component=global_component['name'])
+            update_component_contacts(global_component, component_contacts)
+
         if args.json:
-            print json.dumps(list(global_components))
+            print json.dumps(global_components)
             return
 
         if global_components:
```
ycheng-aa, line 72: If not args.json, the contact information won't be used at all, but above already got them. If global_components already is a list, 'if global_components" looks unnecessary.

pdc_client/plugins/component.py
```diff
@@ -181,8 +188,16 @@ def list_release_components(self, args):
...
+                                           component=release_component['name'],
+                                           release=self._get_release_id(release_component))
+            update_component_contacts(release_component, component_contacts)
+
         if args.json:
-            print json.dumps(list(release_components))
+            print json.dumps(release_components)
             return
 
         if release_components:
```
ycheng-aa, line 201: Same as for global component.

xychu:
I vote for not including the contacts by default unless they are required.
And just realize that with the replacement, the old contact related filters('contact-role' and 'email') in list op will make not that much sense.

xychu:
After we remove the old contacts, now there's no contacts info in component list op result.

ycheng-aa:
Looks good.

PR #2, opened 2015-11-18
Fix pdc client release clone.

xychu:
Make '--include-trees' args to form a list.
JIRA: PDC-1149


lubomir:
This does fix the problem, but I'm not sure it is the best possible fix. The usage is `--include-trees X --include-trees Y`. This is inconsistent with how for example RPM dependencies are handled, which expect the arguments to be `--requires X Y`. I don't know which way is better, but it should be the same everywhere.

xychu:
@lubomir forgot about the `--requires`, and I think `nargs` is better(fewer input) for user and consistent.
Have updated.

lubomir:
:+1:

ycheng-aa:
Looks good.

PR #3, opened 2015-11-18
Make auto-completion more user-friendly.

zkl94:
JIRA: PDC-1161

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -78,6 +78,18 @@ performing the network requests manually.
 This is much more user friendly user interface. A single invocation can
 perform multiple requests depending on what subcommand you used.
 
+The `pdc` client supports Bash completion if argcomplete Python package is installed.
+
+If you installed client from rpm package, the completion file `pdc.bash` has been
+installed to `/etc/bash_completion.d/`.
+
+For developers or users who try to run `pdc` from source, to enable completion,
+run this in your terminal (assuming pdc is somewhere on path).
+
+    eval "$(register-python-argcomplete pdc)"
+
+or put `pdc.bash` to `/etc/bash_completion.d/`.
+
 ## Python API
 
 When writing a client code interfacing with PDC server, you might find
@@ -90,6 +102,7 @@ include
 - [requests](http://docs.python-requests.org/en/latest/)
 - [requests-kerberos](https://github.com/requests/requests-kerberos/)
 - [beanbag](http://beanbag.readthedocs.org/en/latest/)
+- (Optional)[argcomplete](http://argcomplete.readthedocs.org/en/latest/_modules/argcomplete.html)
 
 Please find more details at： [`PDCClient`](pdc_client/__init__.py#L71)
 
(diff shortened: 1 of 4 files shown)


zkl94:
reference of pdc repo: release-engineering/product-definition-center#213

I think of a third solution where we can put `python-argcomplete.sh` file in bash completion scripts directory during package installation rather than put completion file of pdc specifically in bash completion scripts directory. So I think there are totally three plans we have now: 
1. put pdc completion file (the result of `eval "$(register-python-argcomplete pdc)"`) in bash completion scripts directory (which is always `/etc/bash_completion.d/` for CentOS/RHEL)
2. let user run `eval "$(register-python-argcomplete pdc)"` when they need it
3. put `python-argcomplete.sh` (which is what `activate-global-python-argcomplete` puts in bash completion scripts directory); and developers should leave a `PYTHON_ARGCOMPLETE_OK` at the head of executables ready for auto-completion.

I have taken three factors in account: `user-friendliness`, `extensibility` and `support for shells`. So here is my analysis:

#### user-friendliness

by saying user-friendliness, I mean the steps users needed to do to make auto-completion work for pdc.
1. for plan 1, there will be two steps users (rather than one): `install bash-completion using yum` and `install python-argcomplete using yum or pip`. bash-completion functionality is not automatically enabled for CentOS/RHEL and must be enabled manually.
2. for plan 2, there will be two steps for users: `install python-argcomplete using yum or pip` and `put the eval command argcomplete suggests in ~/.bashrc`
3. for plan 3, there will be two steps for users which is the same with the first plan.

Although they all require 2 steps to make it work, plan 1 and plan 3 outweight plan 2 because in the plan 2 editing of ~/.bashrc file is needed. plan 1 and plan 3 wins.

#### extensibility
1. plan 1: for developers, when a new executable is to be added to pdc (if any), he will have to add a completion file for that exec and edit the spec file. This produces completion files scattered in pdc_client; for users, there is nothing to do when pdc upgrades.
2. plan 2: for developers, there is nothing to do other than write the new exec; for users, they have to manually add that to ~/.bashrc
3. plan 3: for developers ,there is nothing needed to do other than add a `PYTHON_ARGCOMPLETE_OK` at the top of the exec; for dev, there is nothing needed to do.

So, as for extensibility, plan 3 wins.

#### support for shells
1. plan 1 only supports bash
2. plan 2 support bash and zsh although there seems to be a different mechanism for zsh to do auto-complete
3. plan 3 only support bash

so, as for support for shells, plan 2 wins. But shells support is not of primary concern.

plan 3 is the same with plan 1 in user-friendliness and beat plan 1 and plan 2 in extensibility. I think it is a workable plan.

xychu:
Any better suggestions, if no for now, how about merge this first?

lubomir:
Personally I would go with option 1. I'm not sure the extra magic of having it work with every python script is worth it and making the completion explicit seems like a good idea. 

It would be a good idea to look at how other projects handle this.

[Fedpkg](https://pagure.io/fedpkg) ships with `fedpkg.bash`, but this is hand-written, package specific script. So does [dnf](https://github.com/rpm-software-management/dnf/blob/master/etc/bash_completion.d/dnf-completion.bash). Finding examples that use `argcomplete` would be better, though.

xychu:
Update this to use the plan 1.
Add `argcomplete` as optional(devel) requirement,
install `pdc.bash` to `/etc/bash_completion.d/`,
and update README to include guide about  bash completion.

simozhan:
:+1:

PR #6, opened 2015-11-19
Update globle-component with its name instead of its id

bliuredhat:
When using pdc command to update globale-component,
we change to use global-component's name instead of
its id as Input Parameter.

JIRA: PDC-1157


ycheng-aa:
Flake8 check fail. Other is good.

lubomir:
:+1:

PR #10, opened 2015-11-19
Fix spec URL.

ralphbean:


pdc-client.spec
```diff
@@ -7,7 +7,7 @@ Release:        1%{?dist}
 Summary:        Console client for interacting with Product Definition Center
 Group:          Development/Libraries
 License:        MIT
-URL:            https://github.com/release-engineering/product-definition-center
+URL:            https://github.com/release-engineering/pdc-client
```
xychu: Thanks @ralphbean and the github org also changed, so the right url will be: https://github.com/product-definition-center/pdc-client

ralphbean: Doh!  I didn't notice.  97ca00d should do it, @xychu.

lubomir:
Merged as f9ab1a1d734389b4627cc1608d699d1152dadb9c.

PR #12, opened 2015-11-26
Release-component info should take release and component

bliuredhat:
Release-component info should take release and component
instead of ID.

JIRA: PDC-1215


xychu:
LGTM, and this also needed for global component, I think, so maybe you could include it as another commit in this pull request.

bliuredhat:
@xychu , yes, I think so for global component.

bliuredhat:
@xychu, ycheng will do the "global component info", this commit I will merge it to master branch.

PR #16, opened 2015-11-30
Enable page_size in new pdc client

bliuredhat:
JIRA: PDC-1158

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -7,6 +7,7 @@
 import itertools
 import json
 import sys
+import exceptions
 
 import beanbag
 import requests
@@ -41,21 +42,10 @@ def read_config_file(server_alias):
 
 
 def get_paged(res, **kwargs):
-    """
-    This call is equivalent to ``res(**kwargs)``, only it retrieves all pages
-    and returns the results joined into a single iterable. The advantage over
-    retrieving everything at once is that the result can be consumed
-    immediately.
+    """ will be abandon in next release"""
+    exceptions.PendingDeprecationWarning("""Warning: This method have been removed into PDCClient class;
+    and then will be deleted in next release.""")
 
-    :param res:     what resource to connect to
-    :param kwargs:  filters to be used
-
-    ::
-
-        # Example: Iterate over all active releases
-        for release in get_paged(client['releases']._, active=True):
-            ...
-    """
     def worker():
         kwargs['page'] = 1
         while True:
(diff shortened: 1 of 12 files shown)


simozhan:
it seems the test failed.

bliuredhat:
Hi, all( @simozhan, @lubomir ,@xychu, @ycheng); 'Enable page_size' is not available for all methods, maybe just for "list"method  and when using extract_arguments() function. Can you give me some more perfect ways to edit this patch?  I think that we don't need to set page_size for ["create" , "update"] methods.
The position of page_size in pdc command as followed:
$   pdc  [-s server]  [page_size   NUM]  command

lubomir:
Alternative way would be to do this change in `get_paged` function. It would probably need to move to `PDCClient` class. I think most of the paged requests go through that.

xychu:
I agree with @lubomir.
It makes more sense that `page_size` get handled and passed to inside functions in `get_paged`.

bliuredhat:
@lubomir  and @xychu , Thanks and I have done it.

lubomir:
Would the type conversion be easier if done with argument parser?
https://docs.python.org/2/library/argparse.html#type

simozhan:
As you move get_paged() to PDCClient, i think you need to update README.markdown as well.

lubomir:
The old function should actually not be deleted: the docstring should be updated to mark it as deprecated. It should only be removed in next release.

bliuredhat:
@lubomir ; yes , your ideal is good; do the command like:
 $ pdc  [-s server]  command   --page-size NUM
?

lubomir:
@bliuredhat The `--page-size` argument should be defined only once, so it would have to come before the command: `pdc -s prod --page-size 11 some-command`.

bliuredhat:
@lubomir , yes~, the command work as you show!

xychu:
I think [ArgumentParser type](https://docs.python.org/2/library/argparse.html#type) is worth to try :-)

bliuredhat:
@xychu , yes; I think it's available ~!

pdc_client/__init__.py
```diff
@@ -42,19 +42,8 @@ def read_config_file(server_alias):
...
-    :param res:     what resource to connect to
-    :param kwargs:  filters to be used
-
-    ::
-
-        # Example: Iterate over all active releases
-        for release in get_paged(client['releases']._, active=True):
-            ...
+    Warning: This method have been removed to PDCClient class;
+    and then will be deleted in next release.
```
xychu: The docstring will not be rendered, so maybe we could add one [PendingDeprecationWarning](https://docs.python.org/2.7/library/exceptions.html#exceptions.PendingDeprecationWarning) as well.

PR #19, opened 2015-12-08
Add help message for 'active' filter.

xychu:
JIRA: PDC-1219

### DIFF
--- a/pdc_client/plugins/component.py
+++ b/pdc_client/plugins/component.py
@@ -145,7 +145,12 @@ class ReleaseComponentPlugin(PDCClientPlugin):
 
         list_parser = self.add_action('list', help='list all release components')
         self.add_include_inactive_release_argument(list_parser)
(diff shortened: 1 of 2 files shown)


lubomir:
Alternative solution would be to have two mutually exclusive options `--active` and `--inactive`. Your solution will work too, I guess.

xychu:
@lubomir, the `--active` and `--inactive` are my first try, it's more direct and neat.
If there is no need to make CLI backward compatible(main reason of my approach),
I'll change this to mutually exclusive options.

lubomir:
For the CLI interface, I _think_ this change would be acceptable.

ycheng-aa:
Looks good.

simozhan:
:+1:

xychu:
Changed to use mutually exclusive group(`--active` and --inactive`).

bliuredhat:
looks good for me

PR #20, opened 2015-12-08
Make mocked endpoints possibly callable.

ralphbean:
This allows you to mock an endpoint with a function that raises a
`beanbag.bbexcept.BeanBagException` so you can test for PDC
failures and other fun stuff.


sochotnicky:
Seems like this should be useful for our QE as well. :+1:

lubomir:
Merged as 8ba2587.

PR #23, opened 2015-12-14
Pypi setup

sochotnicky:
This will need tweaks later I am sure...but might do for a start. If nothing else then integrating with tito is probably needed

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,5 @@
+recursive-include requirements *
+recursive-include docs * 
+recursive-include tests *
+include LICENSE README.markdown tox.ini pdc.bash
+global-exclude  *.pyc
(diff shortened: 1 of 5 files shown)


lubomir:
With regards to the duplicated version, I think the best way forward would be to define the version in `setup.py` only, and add a `get_version` function to the library. The function would try to get version by importing `pkg_resources` and looking for installed module. If that fails (because package is not installed), it could fall back to `git describe`. This way, when running from installed package, it would return the correct version, and in development it would return something identifying the code as well.

sochotnicky:
@lubomir I am fairly certain that will fail because it will try to import dependencies _before_ installing them

edit: oh wait...nevermind :-)

xychu:
Looks good to me.
BTW, I updated the Coverage failure threshold and rerun the job.

lubomir:
Ok then, this version is only really shown in development. Let's use it like this. One minor nitpick: the hash has a trailing new-line.

sochotnicky:
I guess a variant of this has been merged. Closing

PR #30, opened 2016-01-05
Add support for compose-tree-locations.

simozhan:
JIRA: PDC-1259


bliuredhat:
I agree with @ycheng-aa 's ideal, because it keep the result tidily.
By the way; {0:<50} can change to {0:<20} or {0:<30}; the current status is too large, and need two lines to show the head result in terminal.

PR #31, opened 2016-01-07
Add header for build image in new pdc client output

ycheng-aa:
JIRA: 1277

### DIFF
--- a/pdc_client/plugins/build_images.py
+++ b/pdc_client/plugins/build_images.py
@@ -39,7 +39,9 @@ class BuildImagePlugin(PDCClientPlugin):
     def _print_build_image_list(self, build_images, with_md5=False):
         fmt = '{image_id}'
         if with_md5:
-            fmt += ' {md5}'
+            fmt = '{image_id:50}{md5}'
+        print fmt.format(image_id='Image-ID', md5='MD5')
+        print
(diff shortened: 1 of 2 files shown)


pdc_client/plugins/build_images.py
```diff
@@ -39,7 +39,9 @@ def register(self):
     def _print_build_image_list(self, build_images, with_md5=False):
         fmt = '{image_id}'
         if with_md5:
-            fmt += ' {md5}'
+            fmt = '{image_id:50}{md5}'
+        print fmt.format(image_id='Image ID', md5='MD5')
+        print
```
bliuredhat, line 44: why need to add a blank line ?

simozhan:
:+1:

pdc_client/plugins/build_images.py
ycheng-aa, line 44: For me it looks more clear. It also displayed like this way in compose-image-rtt-tests, compose-tree-locations. Not sure if this way is real better. @simozhan what do you think?

simozhan, line 44: i have the same idea. from the output, it is more clear to add a new line between the results and the title.

lubomir:
Looks good to me.

PR #36, opened 2016-01-20
Let pdc client handle pdc warning header

ycheng-aa:
If server return 200 OK with header containing
'pdc-warning' field, display the warning to
customer in the top leverl of the json result.

JIRA: PDC-1283

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -27,6 +27,8 @@ CONFIG_URL_KEY_NAME = 'host'
 CONFIG_INSECURE_KEY_NAME = 'insecure'
 CONFIG_DEVELOP_KEY_NAME = 'develop'
 CONFIG_TOKEN_KEY_NAME = 'token'
+# PDC warning field in response header
+PDC_WARNING_HEADER_NAME = 'pdc-warning'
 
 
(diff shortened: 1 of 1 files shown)


lubomir:
I think this is not the best approach: modifying the response could potentially cause problems for consumers, and when `pdc` is updated to actually display the warning to the user, it would need to be implemented in every subcommand. I think it would be easier to just print the warning to `sys.stderr` in the `decode` method.

The ultimate solution might be to pass an optional logger to `PDCClient`. The warning would than use that, and it would also allow us to add some debugging messages inside the client class if needed. For the `pdc` client we already have a logger we could reuse, for `pdc_client` we would need to create one. Maybe [getting a child logger](https://docs.python.org/2/library/logging.html#logging.Logger.getChild) would be a good choice.

ycheng-aa:
Thanks. I also thought it may break customers' script, but I was not sure how to present it. I just changed it to write to sys.stderr. First make it be able to present the warning.

lubomir:
:+1:

simozhan:
looks good to me. and maybe in the future we can put it in the logger, and provide some option  for the user to see verbose logs

PR #37, opened 2016-01-26
Return error info when input irregular or illegal para.

bliuredhat:
JIRA: PDC-998

### DIFF
--- a/bin/pdc_client
+++ b/bin/pdc_client
@@ -51,7 +51,7 @@ class RequestMethod(object):
     @debug_request
     def get(self, data):
         if isinstance(data, list):
-            print 'Parameter %s is error format %s, please input right format.' % (data, type(data))
+            print 'For GET request, the data must be a JSON object, not a list.'
             sys.exit(1)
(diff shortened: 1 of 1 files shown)


bliuredhat:
@lubomir , @simozhan  and @ycheng-aa ; please review this patch.

bin/pdc_client
```diff
@@ -50,6 +50,9 @@ class RequestMethod(object):
 
     @debug_request
     def get(self, data):
+        if isinstance(data, list):
+            print 'Parameter %s is error format %s, please input right format.' % (data, type(data))
```
ycheng-aa: Should the data be a dict? If so, maybe the condition could be 
if not isinstance(data, dict):
    .........................

Should we also point out what is the correct format in error message?

ycheng-aa:
I also prefer the commit message more specific rather than the PDC number.

bin/pdc_client
bliuredhat: yes,  your ideal is right.

PR #39, opened 2016-01-28
Update the error info

bliuredhat:
The pdc_client command to return "Client error: ", this patch
will remove these info

JIRA: PDC-998

### DIFF
--- a/bin/pdc_client
+++ b/bin/pdc_client
@@ -50,7 +50,7 @@ class RequestMethod(object):
 
     @debug_request
     def get(self, data):
-        if isinstance(data, list):
+        if not isinstance(data, dict):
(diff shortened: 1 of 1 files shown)


bliuredhat:
@lubomir, @simozhan , @ycheng-aa ; please review this patch.

lubomir:
:+1:

bliuredhat:
@lubomir; I consider that if we remove these code, it's bad for  code debug.

lubomir:
I'm not sure how much help it is debugging, but we can keep. In that case there should be an `if` condition for [`SystemExit`](https://docs.python.org/2/library/exceptions.html#exceptions.SystemExit) exception and not print anything for it.

ycheng-aa:
:+1:

PR #41, opened 2016-02-09
Rework spec file according to packaging guidelines

sochotnicky:
There's couple of changes:
- dep on python-beanbag instead of beanbag
- python3 subpackage (disabled due to deps for now)
- split into python2/3 subpackages
- use of Fedora py_build macros


sochotnicky:
https://fedoraproject.org/wiki/Packaging:Python

lubomir:
Merged as b5447ba.

PR #43, opened 2016-02-28
Upadte the get_version function to get the version value

bliuredhat:
When we install the pdc client from brew, we can't get the correct
version value. So add the version value by manual operation.

JIRA: PDC-1375


ycheng-aa:
Looks good for me.

simozhan:
:+1:

lubomir:
I don't like this fix. I think the code that is there should work correctly and the only reason why it fails is because the `.egg-info` directory is not installed. That is an error in packaging and should be fixed there.

bliuredhat:
@lubomir ; your ideal is reasonable~! BTW,  I want to consult you how can we fix this error; the SPEC file is [LINK](https://docs.google.com/a/redhat.com/document/d/1tUaXYgMnTFmL7PdPmizbepKkJZN4tatw9JQ0ec92XMw/edit?usp=sharing)

bliuredhat:
Hi,all guys; This patch will be removed and do not merge to master branch; BTW, thanks for lubomir's help.

PR #44, opened 2016-02-28
Add docs about how to deploy the project both in Brew and in Copr

bliuredhat:
This patch add the doc to show the steps how to deploy pdc-client both in Brew and in Copr.

JIRA: PDC-1349


bliuredhat:
@sochotnicky  and @lubomir ; the internal Copr have been removed; So I am not sure whether we need add the steps about how to deploy release in Copr?

bliuredhat:
@sochotnicky  and @lubomir ; This patch is to show the steps how to deploy the pdc client  package in Brew and Internal Copr. But I consider that these steps are  just access to internal users, especially PDC team-members; not available to external users. Do we need to share these steps  in Github ?

PR #48, opened 2016-04-13
Modify compose-tree-locations in client because API url changed.

ycheng-aa:
Because scheme is added to compose-tree-locations API url,
pdc client is changed correspondingly in info/update/delete.

JIRA: PDC-1445


simozhan:
looks good to me.

bliuredhat:
:+1:

PR #51, opened 2016-05-19
fix version check for the requests module

nphilipp:
Simply comparing strings, '2.10.0' < '2.4.0'. Split the version string,
and compare each component as an integer number.


lubomir:
:+1:

ycheng-aa:
:+1:

PR #52, opened 2016-06-06
Add field 'subvariant' to image sub-command.

ycheng-aa:
JIRA: PDC-1521


simozhan:
I tried to list an existing image, it seems no Subvariant field, does that mean Subvariant field is for future imported images

ycheng-aa:
@simozhan  You could use --json to see full information. Without --json, it's just part of the information.

lubomir:
The field was added only recently, and as far as I know no internal product is using it. It is used in Fedora.

simozhan:
ok.  :+1:

bliuredhat:
looks good for me

PR #53, opened 2016-06-12
Change configuration files for pdc-client

bliuredhat:
The configuration for client should be in a dropdir (/etc/pdc.d/)
so that multiple config files can be installed at the same time.
1. Load a config dict from each file.
2. If two dicts contain the same key, report an error (the error
message should include the key and all file names defining it)
3. Merge all these dicts.
4. Continue as before: load user config file and apply changes from it

JIRA: PDC-1563

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -21,7 +21,7 @@ from pdc_client import monkey_patch
 
 monkey_patch.monkey_patch_kerberos()
 
-GLOBAL_CONFIG_FILE = '/etc/pdc/client_config.json'
+GLOBAL_CONFIG_DIR = '/etc/pdc.d/'
 USER_SPECIFIC_CONFIG_FILE = expanduser('~/.config/pdc/client_config.json')
 CONFIG_URL_KEY_NAME = 'host'
 CONFIG_INSECURE_KEY_NAME = 'insecure'
@@ -54,6 +54,25 @@ def get_version():
 __version__ = get_version()
 
 
+def _read_dir(file_path):
+    # get all json files in /etc/pdc.d/ directory and merge them
+    data = {}
+    if isdir(file_path):
+        files_list = os.listdir(file_path)
+        for file_name in files_list:
+            if file_name.endswith('.json'):
+                file_abspath = os.path.join(file_path, file_name)
+                with open(file_abspath, 'r') as config_file:
+                    config_dict = json.load(config_file)
+                    same_key = set(data.keys()) & set(config_dict.keys())
+                    if same_key:
+                        print "Error: '%s' keys existed in both %s config files" % (same_key, files_list)
+                        sys.exit(1)
+                    else:
+                        data.update(config_dict)
+    return data
+
+
 def _read_file(file_path):
     data = {}
     if isfile(file_path):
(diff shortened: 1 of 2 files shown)


simozhan:
do we want to restrict to these tow files redhat.json and fedora.json, can we add other file like my.json.
and is it ok to expose redhat.json in github? is it ok to search all json files in pdc.d and merge them.
by the way, the coverage decreased, so maybe we need to add some test cases.

lubomir:
There should be no hard coded filenames. The client should load all json files from the directory and merge them.

bliuredhat:
@lubomir , do you means that get all json files in  /etc/pdc.d/ directory and merge them;  if there are same keys  in them, raise error info?

lubomir:
@bliuredhat yes, exactly that.

pdc_client/__init__.py
```diff
@@ -63,7 +64,17 @@ def _read_file(file_path):
 
 
 def read_config_file(server_alias):
-    result = _read_file(GLOBAL_CONFIG_FILE).get(server_alias, {})
+    redhat_config = _read_file(GLOBAL_CONFIG_FILE_REDHAT)
+    fedora_config = _read_file(GLOBAL_CONFIG_FILE_FEDORA)
+    same_key = list(set(redhat_config) & set(fedora_config))
```
ycheng-aa: Seems list() is unnecessary here.

pdc_client/__init__.py
```diff
@@ -56,14 +56,23 @@ def get_version():
...
+                    sys.exit(1)
+                else:
+                    data.update(config_dict)
     return data
 
 
 def read_config_file(server_alias):
-    result = _read_file(GLOBAL_CONFIG_FILE).get(server_alias, {})
+    result = _read_file(GLOBAL_CONFIG_DIR).get(server_alias, {})
     result.update(_read_file(USER_SPECIFIC_CONFIG_FILE).get(server_alias, {}))
```
simozhan, line 86: USER_SPECIFIC_CONFIG_FILE is a file, {} will be returned from _read_file, right?

bliuredhat, line 86: yes

bliuredhat:
@simozhan; I think this patch don't need to add a test case.

pdc_client/__init__.py
```diff
@@ -56,14 +56,23 @@ def get_version():
 
 def _read_file(file_path):
     data = {}
-    if isfile(file_path):
-        with open(file_path, 'r') as config_file:
-            data = json.load(config_file)
+    if isdir(file_path):
+        files_list = os.listdir(file_path)
+        for file_name in files_list:
```
ycheng-aa, line 62: It looks in previous discussion, it is to check all `.json` files, right?

bliuredhat, line 62: yes, check all .json file.

ycheng-aa:
:+1:

lubomir:
As far as functionality goes, this looks good. Since the `_read_file` function is getting quite long and complicated, maybe adding a comment explaining what it is supposed to do might be helpful in the future. But that is up to you.

simozhan:
looks good to me

PR #55, opened 2016-06-23
Move plugins outside of python_sitelib

bliuredhat:
Move all plugins in upstream to /usr/share/pdc-client directory.

JIRA: PDC-1567


ycheng-aa:
Looks good.

So the plugin files will directly under `pdc-client` directory instead of `plugins` directory, right?

simozhan:
LGTM

PR #57, opened 2016-06-27
Allow specifying plugins in the config file

simozhan:
JIRA: PDC-1564

### DIFF
--- a/pdc_client/runner.py
+++ b/pdc_client/runner.py
@@ -37,9 +37,26 @@ from pdc_client.utils import pretty_print
 
 # A list of paths to directories where plugins should be loaded from.
 # The purpose of the plugins is to extend the default behaviour.
-PLUGIN_DIRS = [
-    os.path.join(os.path.dirname(__file__), 'plugins')
+LOCAL_DIR = os.path.join(os.path.dirname(__file__), 'plugins')
+INSTALLED_DIR = os.path.join('/usr/share/pdc-client', 'plugins')
+PLUGIN_DIRS = [LOCAL_DIR] if not os.path.exists(INSTALLED_DIR) else [INSTALLED_DIR]
+
+DEFAULT_PLUGINS = [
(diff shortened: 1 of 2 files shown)


lubomir:
I didn't test it, but the code looks very good. :+1:

pdc_client/runner.py
```diff
@@ -49,10 +66,22 @@ def __init__(self):
...
+        if not server:
+            raise TypeError('Server must be specified')
+        config = pdc_client.read_config_file(server)
+        if config:
+            plugins = config[CONFIG_PLUGINS_KEY_NAME]
+            if not plugins:
+                plugins = DEFAULT_PLUGINS
+        else:
+            plugins = DEFAULT_PLUGINS
+
```
ycheng-aa, line 77: Seems from line 74 it could be:

```
plugins = DEFAULT_PLUGINS
if config and config[CONFIG_PLUGINS_KEY_NAME]:
     plugins = config[CONFIG_PLUGINS_KEY_NAME]
```

to make it a little shorter.

pdc_client/runner.py
```diff
@@ -49,10 +66,22 @@ def __init__(self):
...
         for dir in PLUGIN_DIRS:
             self.logger.debug('Loading plugins from {0}'.format(dir))
             for name in os.listdir(dir):
-                if not name.endswith('.py'):
+                if not name.endswith('.py') or name not in plugins:
```
ycheng-aa, line 81: It's a minor thing. If use `in` here, it's a litter better to set  `plugins` as a set instead of a list. `in` operation for a set is O(1) complexity (https://wiki.python.org/moin/TimeComplexity). But not big deal for this small list.

ycheng-aa:
:+1:

pdc_client/runner.py
simozhan, line 81: @ycheng-aa Interesting. will think about this when met large list. :)

PR #58, opened 2016-06-30
Fix failure with requests-kerberos 0.9+ and Python 3.

matysek:
Removing monkey_patch.py makes Python 3 working with requests-kerberos 0.8, 0.9 and 0.10. Python 2 is still failing for some reason with requests-kerberos 0.9+.


ycheng-aa:
Will this patch affect pdc_client in python 2.x if the old behavior is removed?

matysek:
@ycheng-aa no, the behavior is still same with python 2.x and kerberos-requests 0.7 and 0.8.

ycheng-aa:
OK, then. :+1:

simozhan:
:+1:

PR #60, opened 2016-07-05
Fix bugs when no plugins input or wrong input

simozhan:
JIRA: PDC-1564

### DIFF
--- a/pdc_client/runner.py
+++ b/pdc_client/runner.py
@@ -72,8 +72,10 @@ class Runner(object):
             raise TypeError('Server must be specified')
         config = pdc_client.read_config_file(server)
         plugins = DEFAULT_PLUGINS
-        if config and config[CONFIG_PLUGINS_KEY_NAME]:
-            plugins = config[CONFIG_PLUGINS_KEY_NAME]
+        if config and config.get(CONFIG_PLUGINS_KEY_NAME):
+            plugins = config.get(CONFIG_PLUGINS_KEY_NAME)
+        if not isinstance(plugins, list):
+            raise TypeError('Plugins must be a list')
 
         for dir in PLUGIN_DIRS:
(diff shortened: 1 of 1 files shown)


pdc_client/runner.py
```diff
@@ -72,8 +72,13 @@ def load_plugins(self):
             raise TypeError('Server must be specified')
         config = pdc_client.read_config_file(server)
         plugins = DEFAULT_PLUGINS
-        if config and config[CONFIG_PLUGINS_KEY_NAME]:
-            plugins = config[CONFIG_PLUGINS_KEY_NAME]
```
ycheng-aa, line 76: Looks like change condition to `if config and config.get(CONFIG_PLUGINS_KEY_NAME):` could work as you expected.

bliuredhat:
:+1:

pdc_client/runner.py
```diff
@@ -72,8 +72,13 @@ def load_plugins(self):
...
+        if config:
+            try:
+                plugins = config[CONFIG_PLUGINS_KEY_NAME]
+            except KeyError:
+                pass
+        if not isinstance(plugins, list):
+            raise TypeError('Plugins must be a list')
```
ycheng-aa, line 78: So tuple and set is not allowed. If it can be specified by user, should we better impose less restriction on the format?

pdc_client/runner.py
simozhan, line 76: yes. fix it

pdc_client/runner.py
simozhan, line 78: for the json file, it can be a dict, but the setting is "plugins": ["image.py", "rpm.py"...], i think list is the proper format. and for set and tuple, it is a python type, i don't know if there is same idea in json.

PR #65, opened 2016-07-22
Update the spec file for pdc client

bliuredhat:
Remove test_helpers.py and test_helpers_py3.py in package.

JIRA: PDC-1635

### DIFF
--- a/pdc-client.spec
+++ b/pdc-client.spec
@@ -10,7 +10,7 @@
 
 Name:           pdc-client
 Version:        1.0.0
-Release:        1%{?dist}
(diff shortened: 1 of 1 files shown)


pdc-client.spec
```diff
@@ -174,6 +174,9 @@ EOF
 
 
 %changelog
+* Sun Jul 17 2016 bliu <bliu@redhat.com> 1.0.0-2
+- new package built with tito
```
simozhan: Is it better to add some informative changelog?

simozhan:
LGTM

PR #67, opened 2016-08-02
Disable pagination when page-size is 0 or -1

bliuredhat:
When page-size <= 0,  the pagination will be disabled.

JIRA: PDC-1636


ycheng-aa:
:+1:

simozhan:
If there is no page_size, what is the behavior, it seems " 'None' is a invalid number, please verify it again" will be pop out.
Another thing is when you query data in this way https://pdc.engineering.redhat.com/rest_api/v1/releases/?page_size=-2
it seems it is the same with page_size=0 or -1, so probably we need to make them  consistent.
BTW, the message in print and the comment can be somehow improved :)

bliuredhat:
@simozhan ; thank you, and I submit again, please help to review it.

simozhan:
LGTM

ycheng-aa:
:+1:

PR #71, opened 2016-08-12
Python 3 build

jpopelka:
As stated in https://bugzilla.redhat.com/show_bug.cgi?id=1352472 all dependencies have been ported so you can enable Python 3 build in spec file.

I've tried to build and briefly use `python3-pdc-client` and it looks OK to me.

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -193,9 +193,9 @@ class PDCClient(object):
         if self.page_size is not None:
             kwargs['page_size'] = self.page_size
 
-        if self.page_size <= 0 and self.page_size is not None:
-            # If page_size <= 0, pagination will be disable.
-            return res(**kwargs)
+            if self.page_size <= 0:
+                # If page_size <= 0, pagination will be disable.
+                return res(**kwargs)
 
         def worker():
             kwargs['page'] = 1
--- a/pdc-client.spec
+++ b/pdc-client.spec
@@ -5,12 +5,9 @@
 %{!?python2_sitelib: %global python2_sitelib %(%{__python2} -c "from distutils.sysconfig import get_python_lib; print (get_python_lib())")}
 %endif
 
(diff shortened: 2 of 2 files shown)


jpopelka:
I've added one non spec related commit which fixes a problem I hit after switching to Python 3.

simozhan:
Looks good to me. i am guessing there may be conflict after we merge back pdc-client 1.1.0-1

jpopelka:
So should I remove the '[spec] sync with fedora' commit ?

pdc_client/__init__.py
```diff
@@ -193,9 +193,9 @@ def get_paged(self, res, **kwargs):
         if self.page_size is not None:
             kwargs['page_size'] = self.page_size
 
-        if self.page_size <= 0 and self.page_size is not None:
-            # If page_size <= 0, pagination will be disable.
-            return res(**kwargs)
+            if self.page_size <= 0:
```
bliuredhat, line 196: I not sure this is right, because None<=0 is true. 
As we know, if pdc client command don't add --page-size parameter, the default value of pagination is 20.

jpopelka, line 196: It can't be `None` as I also moved it into the `if self.page_size is not None:` section, see the added white-spaces.

bliuredhat, line 196: +1; yes; you are right.

simozhan:
@jpopelka i guess even removing commit '[spec] sync with fedora', there is still conflict in the changelog. Is it ok to rebase your fix after we merge back code from release branch? It should happen quite soon.

jpopelka:
> Is it ok to rebase your fix after we merge back code from release branch?

sure, no problem

bliuredhat:
@jpopelka ; release branch have merged to master branch.

jpopelka:
rebased

simozhan:
@jpopelka thanks for contributing. Merged.

PR #74, opened 2016-09-14
Update the SPEC file and prepared to build the external package

bliuredhat:
Update the SPEC file and then build the external package of pdc
client in Brew.

JIRA: PDC-1707


pdc-client.spec
```diff
@@ -4,6 +4,9 @@
 %{!?__python2: %global __python2 /usr/bin/python2}
 %{!?python2_sitelib: %global python2_sitelib %(%{__python2} -c "from distutils.sysconfig import get_python_lib; print (get_python_lib())")}
 %endif
+%{!?py2_build: %global py2_build %{expand: CFLAGS="%{optflags}" %{__python2} setup.py %{?py_setup_args} build --executable="%{__python2} -s"}}
```
simozhan, line 6: i guess py2_build is similar to __python2, so it is better to put it in `else...endif`, right?

bliuredhat: yes, it's right.

simozhan:
LGTM

ycheng-aa:
Change looks good. It's better to update commit message about update spec file for what, give more information.  `build the external package of pdc client in Brew` looks not a part of this pull request. Maybe should not be in commit message.

PR #77, opened 2016-11-11
More flexible SSL certificate verification configuration

ahills:
The Python `requests` library has more flexible configuration for SSL certificate verification than the PDC client currently permits. This patch removes the `insecure` configuration option in favor of `ssl_verify` (in the class constructor keyword arguments) and `ssl-verify` (in the JSON configuration format). The value is passed through directly to the `requests` session.

This is particularly useful for validation against a specific certificate, which can be specified as a path. See the [`requests` documentation](http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification) for more information.

### DIFF
--- a/pdc-client.spec
+++ b/pdc-client.spec
@@ -147,7 +147,7 @@ cat > %{buildroot}/%{_sysconfdir}/pdc.d/fedora.json << EOF
     "fedora": {
         "host": "https://pdc.fedoraproject.org/rest_api/v1/",
         "develop": false,
-        "insecure": false
+        "ssl-verify": true
     }
 }
 EOF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -17,7 +17,6 @@ from subprocess import Popen, PIPE
 import beanbag
 import requests
 import requests_kerberos
-import warnings
 
 GLOBAL_CONFIG_DIR = '/etc/pdc.d/'
 USER_SPECIFIC_CONFIG_FILE = expanduser('~/.config/pdc/client_config.json')
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -22,7 +22,7 @@ import warnings
 GLOBAL_CONFIG_DIR = '/etc/pdc.d/'
 USER_SPECIFIC_CONFIG_FILE = expanduser('~/.config/pdc/client_config.json')
 CONFIG_URL_KEY_NAME = 'host'
-CONFIG_INSECURE_KEY_NAME = 'insecure'
+CONFIG_SSL_VERIFY_KEY_NAME = 'ssl-verify'
 CONFIG_DEVELOP_KEY_NAME = 'develop'
 CONFIG_TOKEN_KEY_NAME = 'token'
 # PDC warning field in response header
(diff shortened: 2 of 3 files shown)


pdc_client/__init__.py
```diff
@@ -120,28 +119,19 @@ class to perform requests.
...
             # just access API directly.
             # REQUIRED, OPTIONAL, DISABLED
             self.session.auth = requests_kerberos.HTTPKerberosAuth(
                 mutual_authentication=requests_kerberos.DISABLED)
 
-        if insecure:
-            # turn off for servers with insecure certificates
-            self.session.verify = False
-            # turn off warnings about making insecure calls
-            if [int(x) for x in requests.__version__.split('.')] < [2, 4, 0]:
```
ycheng-aa, line 138: Should the message still be kept?

pdc-client.spec
```diff
@@ -147,7 +147,7 @@ cat > %{buildroot}/%{_sysconfdir}/pdc.d/fedora.json << EOF
     "fedora": {
         "host": "https://pdc.fedoraproject.org/rest_api/v1/",
         "develop": false,
-        "insecure": false
+        "ssl-verify": true
```
ycheng-aa, line 150: @bliuredhat redhat's configuration should also be changed that way, right? It is not related to this patch, but it looks if this package is upgraded, the configuration above will overwrite the existing one. I am not sure if this part should be put in %post part to judge install or upgrade, and in upgrading just change this one item and keep others.

bliuredhat: @ycheng-aa ; yes, you are right; it need to update the config file  in sync.

pdc_client/__init__.py
ahills: I think it's up to each application to decide what to do about the warnings. It's not behavior I expect from a library. If you want, I will add this functionality back to the scripts in `bin`.

pdc-client.spec
ahills: I'm not experienced writing RPM spec files; can any section be written in Python, to make JSON access easier? Maybe it would be best if I modified the source here to accept both values from the configuration, and emit a warning about a deprecated option?

lubomir:
One thing is that no amount of specfile magic will help with config files in `~/.config`. I think it would be nice to print an explicit error when the old option is is used (or print a deprecation warning and somehow still accept the old option).

pdc-client.spec
ycheng-aa: I think @lubomir 's suggestion is easy and compatible to previous version. Could we just accept both `insecure` and `ssl-verify` and convert it internally?

PDC team should change spec file accordingly in internal pdc client spec file. And better change spec file to don't overwrite original pdc client global configuration file in every upgrade. These are not related to this patch itself.

pdc_client/__init__.py
ycheng-aa: Sounds good for me.

ahills:
@ycheng-aa @lubomir 

After a long delay, the `insecure` option is back with a warning. Its behavior is almost the exact opposite of `ssl_verify`, so `ssl_verify = not insecure` is used to implement the old behavior.

A side effect of this particular implementation (intentional) is that the `insecure` option, if present, will override any given `ssl-verify` value. I can also change it to the opposite, if you think that's best.

ycheng-aa (approved):

lubomir (approved):

ahills:
@lubomir @ycheng-aa

I don't have write access; could one of you merge this for me? Thanks!

ycheng-aa:
Done.

PR #78, opened 2016-11-25
Support option --server as well.

simozhan:
JIRA: PDC-1718

### DIFF
--- a/pdc_client/runner.py
+++ b/pdc_client/runner.py
@@ -69,15 +69,22 @@ class Runner(object):
     def load_plugins(self):
         config = None
         server = None
+        idx_s, idx_server = (None, None)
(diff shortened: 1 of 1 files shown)


pdc_client/runner.py
```diff
@@ -69,15 +69,26 @@ def __init__(self):
...
         except ValueError:
             pass
         except IndexError:
             pass
+        try:
+            idx_server = args.index('--server')
+        except ValueError:
+            pass
+        except IndexError:
+            pass
```
ycheng-aa, line 82: This kind of except could write something like:
```
except (ValueError, IndexError) as e:
    pass
```
Same as previous try....except......

simozhan: Sure. Fixed

ycheng-aa (approved):

PR #80, opened 2016-12-05
Bug about ssl verify

bliuredhat:
The value of "insecure" should be related to CONFIG_INSECURE_KEY_NAME
in config file.


ycheng-aa:
Looks good.

ahills:
Good catch!

PR #84, opened 2016-12-26
Fix the problem in repo clone

ycheng-aa:
Value of ome of the command parameters, e.g.
include_service, include_repo_family,
include_content_format and include_content_category
should be a list instead of single
element.
Also add arguments for from release id
and to release id in this case to avoid
treating them as part of a list for other
arguments.

JIRA: PDC-1801


ycheng-aa:
It looks could work with command like `pdc -s qa -k content-delivery-repo clone  --include-content-format iso --release-id-from rhel-7.0 --release-id-to abxw-2357`.

bliuredhat:
+1

simozhan:
+1

PR #85, opened 2017-01-03
Fix the bug about the include-shadow para in repo clone

bliuredhat:
When do repo clone with [--include-shadow | --exclude-shadow]
parameter,the command can't work well.

JIRA:PDC-1817


ycheng-aa:
It's better also repo clone related tests.

bliuredhat:
@ycheng-aa and @simozhan ; please help to review this patch.

tests/repo/tests.py
```diff
@@ -131,3 +131,40 @@ def test_delete(self, api):
...
+                "variant_uid": "Workstation"
+            }
+
+            for x in range(1, 30)
+        ])
+        self._setup_detail(api)
+        with self.expect_output('list.txt'):
+            self.runner.run(['content-delivery-repo', 'clone',
+                             '--release-id-from', 'rhel-7.1-updates',
+                             '--release-id-to', 'rhel-7.1'
```
ycheng-aa, line 157: This is for --include-shadow and --exclude-shadow take effect. Could also add corresponding tests?

simozhan:
i can't run the test successfully. can you help to check it?

tests/repo/tests.py
bliuredhat: fix in https://github.com/product-definition-center/pdc-client/pull/85/commits/d3a39a99703fdf8cf925cf8b9f85394c15e86662

simozhan:
looks good to me

ycheng-aa:
I think list_clone_false.json should have a more meaningful name. It means shadow is false, right? Could it be something like list_clone_shadow_false.json?
Other part looks for me.

ycheng-aa:
Looks good.

PR #86, opened 2017-01-20
Surport SSL cert when swith insecure to false

bliuredhat:
When input the parameters with --ca-cert, the client can
surport it.

JIRA: PDC-1719


pdc_client/__init__.py
```diff
@@ -125,7 +125,10 @@ class to perform requests.
             if insecure is not None:
                 sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
                     CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
-                ssl_verify = not insecure
+                if insecure:
+                    ssl_verify = not insecure
+                else:
+                    ssl_verify = ssl_verify or not insecure
```
ycheng-aa: Why this line is different from 129, it looks if insecure is true we will use insecure value but insecure is false we try to use ssl_verify value first, but if ssl_verify is false then we still use insecure value.

Could you add a little comment to explain the reading configuration logic?

simozhan: I am doubting if ssl_verify can get path from the config file(Line 123) or not. As code in  pdc_client/runner.py#L154(ssl_verify = self.args.ca_cert or not self.args.insecure)
ssl_verify is path/True/False, it will not be None, right? Please correct me, maybe i misunderstand something

bliuredhat:
@ycheng-aa and @simozhan ; I consider when "insecure "=false, the "ssl_verify" ->path/to/CA file. Meanwhile, If the "insecure "=false and "ssl_verify" = None ; change "ssl_verify "=True.

pdc_client/__init__.py
bliuredhat: @simozhan, exactly~!  When "insecure" =False,  expect that "ssl_verify" is the path to CA file.

bliuredhat: @ycheng-aa, I hope ssl_verify is the path to CA file; not only `True/Flase`.  By the way,   ssl_verify can not be `None` , It's more reasonable.

ycheng-aa: You said `not only` True/False. But in this if ssl_verify is False, it'll not take effect, right? It seems refuse to accept ssl_verify being false.

bliuredhat: @ycheng-aa ; I think False will be affect.  If the `insecure` is True, `ssl_verify` will be False. Do you think so?

ycheng-aa: From line 131 if ssl_verify set as fase, it will not take effect and use the value `not insecure`, that should be true here. So ssl_verify will be true in line 131 no matter what ssl_verify is set in cofig. Is it what you expect? If it is, seems a little complicate to do it. Could simple `ssl_verify=true` do the same thing?

bliuredhat: @ycheng-aa ; I undstand your means~! In my opinion, in current status;I consider that the `insecure` is more high priority than `ssl_verify`, do you think so? I have update the patch, which maybe more simple and clear. Please switch to new patch and help  to review it. 
I will send you and @simozhan a email to discuss about that I want to remove "insecure" key and only "ssl_verify" key in client.

pdc_client/__init__.py
```diff
@@ -125,7 +125,10 @@ class to perform requests.
             if insecure is not None:
                 sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
                     CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
-                ssl_verify = not insecure
+                if insecure:
+                    ssl_verify = False
+                elif not insecure and not ssl_verify:
```
simozhan: elif already means insecure is false, right?

bliuredhat: @simozhan , yes.

ycheng-aa: agree

ycheng-aa:
It seems the @simozhan's comment haven't been fixed.

bin/pdc_client
```diff
@@ -169,7 +169,7 @@ if __name__ == "__main__":
                 data[key] = ''
 
     try:
-        ssl_verify = options.ca_cert or not options.insecure
+        ssl_verify = options.ca_cert or options.insecure
```
ycheng-aa: Why is ssl_verify not opposite of insecure?

simozhan: i don't understand why removing not here

pdc_client/runner.py
```diff
@@ -151,7 +151,7 @@ def setup(self):
 
     def run(self, args=None):
         self.args = self.parser.parse_args(args=args)
-        ssl_verify = self.args.ca_cert or not self.args.insecure
+        ssl_verify = self.args.ca_cert or self.args.insecure
```
simozhan: same as above

pdc_client/__init__.py
```diff
@@ -120,16 +120,18 @@ class to perform requests.
...
-                    ssl_verify = True
+            if ssl_verify is None:
+                ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME)
+                insecure = config.get(CONFIG_INSECURE_KEY_NAME)
+                if insecure is not None:
+                    sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
+                        CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
+                    if insecure:
+                        ssl_verify = False
+                    elif not insecure and not ssl_verify:
```
ycheng-aa: As @simozhan suggested, `elif` already  mean `not insecure`.

bin/pdc_client
bliuredhat: @simozhan ; If options.insecure is `None`, it will value `ssl_verify` -> `True`. This is not right, if there is no --insecure in client command line, we hope ssl_verify is None in line 176.

pdc_client/__init__.py
```diff
@@ -120,16 +120,18 @@ class to perform requests.
...
+                        ssl_verify = True
+            elif ssl_verify is True:
+                ssl_verify = False
```
ycheng-aa: So `insecure` is false mean `ssl_verify` is false here. Why ssl_verify should be the same as insecure value here?

bin/pdc_client
simozhan: ssl_verify is a bit confused. i guess this variable means ca_cert_or_insecure, its value is None/pathtocert/True
is that correct? And i think you should add some comments as well.

pdc_client/runner.py
simozhan: the same as above

pdc_client/__init__.py
```diff
@@ -120,12 +120,18 @@ class to perform requests.
...
-            ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME) if ssl_verify is None else ssl_verify
-            insecure = config.get(CONFIG_INSECURE_KEY_NAME)
-            if insecure is not None:
-                sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
-                    CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
-                ssl_verify = not insecure
+            if ssl_verify is None:
+                ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME)
+                insecure = config.get(CONFIG_INSECURE_KEY_NAME)
+                if insecure is not None:
```
simozhan: I think we should first give some warning for end user if ssl_verify and insecure has the same value, right?

pdc_client/__init__.py
```diff
@@ -120,12 +120,18 @@ class to perform requests.
...
                 sys.exit(1)
-            ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME) if ssl_verify is None else ssl_verify
-            insecure = config.get(CONFIG_INSECURE_KEY_NAME)
-            if insecure is not None:
-                sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
-                    CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
-                ssl_verify = not insecure
+            if ssl_verify is None:
+                ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME)
+                insecure = config.get(CONFIG_INSECURE_KEY_NAME)
```
simozhan: I think we should first give some warning for end user if ssl_verify and insecure has the same value, right?
Or we want to tell user which one is in use?

pdc_client/__init__.py
```diff
@@ -120,12 +120,18 @@ class to perform requests.
...
+                if insecure is not None:
+                    sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
+                        CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
+                    if insecure:
```
simozhan: if there is waring before hand, we can simply return ssl_verify or not insecure, right?

pdc_client/__init__.py
```diff
@@ -120,12 +120,18 @@ class to perform requests.
             except KeyError:
                 print("'%s' must be specified in configuration file." % CONFIG_URL_KEY_NAME)
                 sys.exit(1)
-            ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME) if ssl_verify is None else ssl_verify
-            insecure = config.get(CONFIG_INSECURE_KEY_NAME)
-            if insecure is not None:
-                sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
-                    CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
-                ssl_verify = not insecure
+            if ssl_verify is None:
```
simozhan: as above, i think ssl_verify should be something like ca_cert_or_insecure

pdc_client/__init__.py
```diff
@@ -120,12 +120,18 @@ class to perform requests.
...
+                ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME)
+                insecure = config.get(CONFIG_INSECURE_KEY_NAME)
+                if insecure is not None:
+                    sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
+                        CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
+                    if insecure:
+                        ssl_verify = False
+                    elif not insecure and not ssl_verify:
+                        ssl_verify = True
+            elif ssl_verify is True:
```
simozhan: as above, i think ssl_verify should be something like ca_cert_or_insecure

pdc_client/__init__.py
```diff
@@ -120,12 +120,18 @@ class to perform requests.
...
+                ssl_verify = False
```
simozhan: then there should be:
else:
    ssl_verify=ca_cert_or_insecure

bin/pdc_client
simozhan: @ycheng-aa ssl_verify is a bit confused. i guess this variable means ca_cert_or_insecure, its value is None/pathtocert/True
is that correct? @bliuredhat

bliuredhat:
@ycheng-aa  and @simozhan ; It maybe more clear that change ssl_verify to ca_cert_or_insecure in some code.

pdc_client/__init__.py
```diff
@@ -120,12 +120,20 @@ class to perform requests.
...
-            ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME) if ssl_verify is None else ssl_verify
-            insecure = config.get(CONFIG_INSECURE_KEY_NAME)
-            if insecure is not None:
-                sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
-                    CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
-                ssl_verify = not insecure
+            if ca_cert_or_insecure is None:
+                ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME)
+                insecure = config.get(CONFIG_INSECURE_KEY_NAME)
+                if insecure is not None:
```
simozhan: if insecure is not None:
    warning
if insecure:
   ssl_verify = False
else:
  ssl_verify = ssl_verify

pdc_client/__init__.py
```diff
@@ -130,8 +130,10 @@ class to perform requests.
                         ssl_verify = False
                     elif not insecure and not ssl_verify:
```
ycheng-aa: Again, don't need `not insecure`.

pdc_client/__init__.py
```diff
@@ -130,8 +130,10 @@ class to perform requests.
...
                         ssl_verify = True
-            elif ssl_verify is True:
+            elif ca_cert_or_insecure is True:
                 ssl_verify = False
+            else:
+                ssl_verify = ca_cert_or_insecure
```
ycheng-aa: If options.ca_cert is `none` and options.insecure is `false` then ssl_verify will be set to `false`. Is it because if no ca_cert then client will not use ssl_verify anyway, even insecure is set to false in command line?

bliuredhat: @ycheng-aa ,There  are 2 questions in your comments.
1. If no ca_cert in client, client also use ssl_verify and can get the value from config file or not insecure.
2. In command line, it can't set insecure as false, only `True` can be set. (https://github.com/product-definition-center/pdc-client/blob/master/pdc_client/runner.py#L130)

ycheng-aa: Got it. Thanks.

pdc_client/__init__.py
```diff
@@ -123,13 +123,18 @@ class to perform requests.
...
                         CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
                     if insecure:
-                        ssl_verify = False
-                    elif not insecure and not ssl_verify:
-                        ssl_verify = True
+                        ssl_verify = not insecure
+                    else:
+                        # ssl_verify can be None here.
+                        ssl_verify = ssl_verify or not insecure
             elif ca_cert_or_insecure is True:
```
ycheng-aa: Maybe add a little comments or/and use meaningful name to designate that ca_cert_or_insecure is from command line, can't be false; insecure is gotten from config file, etc, could make it easier to understand.

ycheng-aa (commented):
Generally looks good to me. Could add tests to test the parameters part? For simple just command line part. I am not sure.

pdc_client/__init__.py
```diff
@@ -120,12 +121,25 @@ class to perform requests.
...
+                    # Give a warning if ssl_verify == insecure in config file
+                    print("In config file, the values of ssl_verify and insecure can't be the same")
+                    sys.exit(1)
+                if insecure is not None:
+                    sys.stderr.write("Warning: '%s' option is deprecated; please use '%s' instead\n" % (
+                        CONFIG_INSECURE_KEY_NAME, CONFIG_SSL_VERIFY_KEY_NAME))
+                    if insecure:
+                        ssl_verify = not insecure
+                    else:
+                        # ssl_verify can be None here.
```
simozhan: this comment is bit confused for me. which ssl_verify you mean can be None?
i guess you mean the second value.
But maybe you can say the value for the first ssl_verify can be a path or not insecure

simozhan:
i agree with @ycheng-aa that maybe we can add some tests here. and another thing is adding more comments

bliuredhat:
@ycheng-aa and @simozhan ; thank for your help to  review this patch. I  don't prefer to add  test case for this patch. For MockAPI class, it's prepared to test API.

ycheng-aa:
Looks good for me.

MockAPI is prepared to test API may be not the reason that don't  add tests. It has possibility to test it without using MockAPI.  Maybe conclusion should be gotten from investigation and if the function should be tested.

bliuredhat:
@ycheng-aa and @simozhan ; I create a JIRA task [PDC-1838](https://projects.engineering.redhat.com/browse/PDC-1838 ) to track your requirement which to add some test case for this patch. Due to I can't find a valid way to solve it right now, and I plan to merge this patch to master branch firstly and do the test case in future.

PR #88, opened 2017-03-01
Fix SSL command line options

ahills:
The behavior of `-k`/`--insecure` and `--ca-cert` is incorrect with no
configuration. The concept of "insecure" should also be restricted to
the command line parser; the client object should only know the value to
hand off to `requests`.

This patch corrects this issue, as well as another issue where command
line arguments would not override configuration values. The basic client
(`pdc_client` script) has been switched from `OptionParser` to
`ArgumentParser` to simplify and keep synchronized with the rich client.

### DIFF
--- a/pdc_client/runner.py
+++ b/pdc_client/runner.py
@@ -159,7 +159,7 @@ class Runner(object):
         elif self.args.ca_cert:
             ssl_verify = self.args.ca_cert
         else:
-            ssl_verify = True
+            ssl_verify = None
         self.client = pdc_client.PDCClient(self.args.server, page_size=self.args.page_size, ssl_verify=ssl_verify)
         try:
             self.args.func(self.args)
--- a/pdc_client/runner.py
+++ b/pdc_client/runner.py
@@ -127,11 +127,12 @@ class Runner(object):
         self.parser.add_argument('-s', '--server', default='stage',
                                  help='API URL or shortcut from config file')
 
-        self.parser.add_argument('-k', '--insecure', action='store_true',
-                                 help='Disable SSL certificate verification')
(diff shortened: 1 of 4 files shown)


pdc_client/runner.py
```diff
@@ -151,12 +152,15 @@ def setup(self):
...
         if self.args.insecure:
             requests.packages.urllib3.disable_warnings(
                 requests.packages.urllib3.exceptions.InsecureRequestWarning)
-        self.client = pdc_client.PDCClient(self.args.server, page_size=self.args.page_size,
-                                           ca_cert_or_insecure=ca_cert_or_insecure)
+            ssl_verify = False
+        elif self.args.ca_cert:
+            ssl_verify = self.args.ca_cert
+        else:
+            ssl_verify = True
```
ycheng-aa: Does it mean if not set related arguments, ssl_verify default is True? If so, in above class PDCClient part,
```
if ssl_verify is None:
     cfg_ssl_verify = config.get(CONFIG_SSL_VERIFY_KEY_NAME)
```
looks will not use argument in configuration file in this condition. Is it expected?

bliuredhat: I agree with @ycheng-aa's idea, I prefer to set `ssl_verify= None`  here.

Other code looks good for me.

ahills: That's a good point. I'll change it to `None`.

ahills:
@ycheng-aa & @bliuredhat, thanks for the feedback. I've made the requested changes.

bliuredhat:
Looks good for me

ycheng-aa:
Looks good.

PR #89, opened 2017-03-10
Add basic product operations

ahills:
Products can be created, updated, retrieved, and listed using the
`create`, `update`, `info`, and `list` commands, respectively. Product
activation/deactivation is not supported by the API. Filtering by
partial names is not supported by the API (only exact matches).


ycheng-aa:
Generally looks good for me. Please add tests for each method then we can see the exactly output. And no tests also cause coverage deceased..

pdc_client/plugins/product.py
```diff
@@ -0,0 +1,145 @@
...
+        if args.internal:
+            filters['internal'] = True
+        elif args.external:
+            filters['internal'] = False
+        if args.name:
+            filters['name'] = args.name
+        if args.short:
+            filters['short'] = args.short
+
+        products = list(self.client.get_paged(self.client.products._, **filters))
```
bliuredhat, line 76: When running this plugin with --page-size optional argument;  the results can't be return as expected.  
Maybe update this line: 
 ` products = self.client.get_paged(self.client.products._, **filters)`
And also need to  update the other codes as sync.

PR #90, opened 2017-03-10
Add basic product version operations

ahills:
Product versions can be created, updated, retrieved, and listed using
the `create`, `update`, `info`, and `list` commands, respectively.
Product version activation/deactivation is not supported by the API.
Filtering by partial names is not supported by the API (only exact
matches).


ycheng-aa:
Generally looks good for me. Please add tests for each method then we can see the exactly output.

lubomir:
> Product version activation/deactivation is not supported by the API.

This is because a products and product versions are active if they have at least one active release. They don't really have this attribute on their own.

pdc_client/plugins/product_version.py
```diff
@@ -0,0 +1,146 @@
...
+        if args.name:
+            filters['name'] = args.name
+        if args.version_id:
+            filters['product_version_id'] = args.version_id
+        if args.short:
+            filters['short'] = args.short
+        if args.version:
+            filters['version'] = args.version
+
+        product_versions = list(self.client.get_paged(self.client['product-versions']._, **filters))
```
bliuredhat, line 75: When running the patch with --page-size optional arg, can't return the results as expected. 
For example: 
$     ./bin/pdc -s dt -k --page-size 10  product-version list
It will return the results as one page,  the --page-size arg may be disabled.
Prefer to update the code: 
 ` product_versions = self.client.get_paged(self.client['product-versions']._, **filters)`
and update other codes as sync.

PR #91, opened 2017-04-28
Fix usability issues in CLI, add missing features.

dmach:


lubomir:
The test are failing because they call a script that no longer exists. The `.travis.yml` file should be updated.

hluk:
@dmach Not sure how much time are you willing to spend on this PR.

We are thinking of cherry-picking separate commits from here, solving conflicts, creating PR request for each JIRA and fixing any further problems. Do you think this would be OK?

xiangge:
@dmach I agree with @hluk , we want to separate these commits.

hluk:
Everything should be merged now except ["Use local plugins..."](https://github.com/product-definition-center/pdc-client/pull/91/commits/64bea69836fd4bdaa321753e1438c44d609a4e81) commit was replaced with f0b60a3.

PR #94, opened 2017-06-30
Pdc client token update

xiangge:
1.Change the new token/secure to chuang cao.
2.Limit the version of Sphinx to 1.2.3 to resolve the python 2.6 build failed issue.
   The root cause of this is adding new typing package dependency in latest Sphinx, 
   and even the oldest typing package can't support 2.6 build.


simozhan:
Looks good to me. Please squash the two commits before merging

simozhan:
:+1:

hluk:
@simozhan Is there any reason to squash the two commits? It looks good to me as is.

hluk:
👍

simozhan:
@hluk yeah. you are right. i don't need to care about the first commit build fail. After resubmit these two commits, only the HEAD one will trigger the Travis CI build. And different features in different commits always make sense.

xiangge:
@simozhan and @hluk , thank you very much for reviewing that. And I agree with you two.
As these two commits solve different things, I will keep them both, and also separated and modified the commit info here.
I will commit them then.

PR #95, opened 2017-07-03
Configuration fixes

hluk:
Fixes name of "ssl-verify" option in documentation (this confused me a lot) and "develop" option (which was ignored before).

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -31,7 +31,7 @@ Other possible keys are:
     If specified, this token will be used for authentication. The client
     will not try to obtain any token from the server.
 
-* `ssl_verify`
+* `ssl-verify`
 
     If set to `false`, server certificate will not be validated. See [Python requests documentation](http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification) for other possible values.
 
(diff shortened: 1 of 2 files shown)


README.markdown
```diff
@@ -31,7 +31,7 @@ Other possible keys are:
     If specified, this token will be used for authentication. The client
     will not try to obtain any token from the server.
 
-* `ssl_verify`
+* `ssl-verify`
```
simozhan, line 34: Good catch. Yeah. this really confuses the user

README.markdown
```diff
@@ -57,7 +57,7 @@ localhost and a production server. :
         "local": {
             "host": "http://localhost:8000/rest_api/v1/",
             "develop": true,
-            "insecure": false
+            "ssl-verify": false
```
simozhan, line 60: Yeah. As we prefer ssl-verify, we should update this.
In our internal repo, the file pdc_client_config.json should also be updated. But that should be in another patch

simozhan:
LGTM :+1:

xiangge:
Looks good to me. 👍
I think this change is more reasonable.

PR #100, opened 2017-07-17
Add "product" command.

chcao55:
JIRA: PDC-1911


pdc_client/plugins/product.py
```diff
@@ -0,0 +1,119 @@
...
+                                 help='show both active and inactive products')
+        list_parser.set_defaults(func=self.product_list)
+
+        # CRUD: UPDATE
+        update_parser = self.add_action('update', help='update an existing product')
+        update_parser.add_argument('short', metavar='SHORT')
+        self.add_product_arguments(update_parser)
+        update_parser.set_defaults(func=self.product_update)
+
+        # CRUD: DELETE
```
bliuredhat, line 50: line 50 and line 51 would be removed.

pdc_client/plugins/product.py
```diff
@@ -0,0 +1,119 @@
...
+        self.add_product_arguments(create_parser, required=True)
+        create_parser.set_defaults(func=self.product_create)
+
+        # CRUD: READ one (info)
+        info_parser = self.add_action('info', help='display details of a product')
+        info_parser.add_argument('short', metavar="SHORT")
+        info_parser.set_defaults(func=self.product_info)
+
+        # CRUD: READ many (list)
+        list_parser = self.add_action('list', help='list all products')
```
bliuredhat, line 37: According to the 'product' API, the parameters of 'name' and 'short'  will be added to filter the results.
Reference: https://pdc.engineering.redhat.com/rest_api/v1/products/#tab_list .

pdc_client/plugins/product.py
```diff
@@ -0,0 +1,119 @@
...
+    def product_create(self, args):
+        data = self.get_product_data(args)
+        self.logger.debug('Creating product with data {0}'.format(data))
+        response = self.client.products._(data)
+        self.product_info(args, response['short'])
+
+    def product_info(self, args, short=None):
+        short = short or args.short
+        product = self.client.products[short]._()
+
```
bliuredhat: The followed codes would be added; and keep the same formate for all plugins.
 if args.json:
       print(json.dumps(product))
       return
I don't test the above code in local evn.

chcao55: @bliuredhat cool, that's correct, thanks for remind that.

pdc_client/plugins/product.py
chcao55: Yes, need add some values here, thanks for point that.

pdc_client/plugins/product.py
chcao55: I think we can keep this comment as a mention, because there is no delete api on server side here.

bliuredhat: yeah, It's okay for me.

bliuredhat:
Looks good for me

pdc_client/plugins/product.py
```diff
@@ -0,0 +1,128 @@
...
+        # CRUD: READ one (info)
+        info_parser = self.add_action('info', help='display details of a product')
+        info_parser.add_argument('short', metavar="SHORT")
+        info_parser.set_defaults(func=self.product_info)
+
+        # CRUD: READ many (list)
+        list_parser = self.add_action('list', help='list all products')
+        list_parser.add_argument('--inactive', action='store_true',
+                                 help='show only inactive products')
+        list_parser.add_argument('--active', action='store_true',
```
simozhan: As we discussed offline, Let's make 'inactive' and 'all' arguments the same with other plugins, like release

xiangge: ok

pdc_client/plugins/product.py
simozhan: I think we don't need to add all the filters here. Maybe the user only wants 'inactive' and 'all'. If we really want other filters, we can create new ticket and add the feature then.

xiangge: @simozhan Yes, I agree with you, this is a good way, I also discussed it with @bliuredhat before.
Now the codes are rolled back, please review again.

simozhan:
LGTM :+1:

pdc_client/plugins/product.py
```diff
@@ -0,0 +1,123 @@
...
+        fmt = '{0:20} {1}'
+        for key, value in self.prep_for_print(product).items():
+            print(fmt.format(key, value))
+
+    def product_list(self, args):
+        filters = {}
+        if args.inactive:
+            filters['active'] = False
+        elif not args.all:
+            filters['active'] = True
```
hluk, line 91:
```py
filters = { 'active': not args.inactive and args.all }
```

hluk (commented):
+1

pdc_client/plugins/product.py
xiangge: I think both ways are okay, so no need to change the original codes, what's your thinking?

hluk: Yes, it's OK to keep it as it is.

hluk (approved):

pdc_client/plugins/product.py
xiangge: Cool!

xiangge:
Close this as patches have been merged.

PR #102, opened 2017-07-20
Add "release-variant" command

xiangge:
Add "release-variant" command.
As the patch needs support 'dest', so also include the below patch which is the same with Jira PDC-1913.
Will commit this 'dest' patch first, after that, will do the rebase work when merge this mainly patch.
plugin_helpers: Allow overriding 'dest' option.

JIRA: PDC-1913


pdc_client/plugins/release_variant.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # CRUD: UPDATE
+        update_parser = self.add_action('update', help='update an existing release_variant')
+        update_parser.add_argument('release')
+        update_parser.add_argument('uid')
+        self.add_release_variant_arguments(update_parser)
+        update_parser.set_defaults(func=self.release_variant_update)
+
+        # CRUD: DELETE
+        # not implemented
```
hluk: Old comment? Looks implemented now.

hluk (commented):
+1

pdc_client/plugins/release_variant.py
xiangge: Yes, another patch here adds the delete function.
Just remove the comment line from that commit.

simozhan:
:+1:

xiangge:
Close this as patches have been merged.

PR #107, opened 2017-07-26
Use local development plugin directory

hluk:
Prefer using local development plugin directory if exists instead of
installed one.

Replaces [this commit on PR](https://github.com/product-definition-center/pdc-client/pull/91/commits/64bea69836fd4bdaa321753e1438c44d609a4e81).


simozhan:
:+1: It seems more and more inconvenient for the developer or qe when they installed a pdc client and also clone source code. I like this fix.

xiangge:
Looks good to me

PR #109, opened 2017-07-28
Remove the arch parameter from option

xiangge:
Remove the arch parameter from option.

JIRA: PDC-2076

### DIFF
--- a/pdc_client/plugins/release_variant.py
+++ b/pdc_client/plugins/release_variant.py
@@ -36,7 +36,7 @@ class ReleaseVariantPlugin(PDCClientPlugin):
 
         # CRUD: READ many (list)
         list_parser = self.add_action('list', help='list all release_variants')
-        self.add_release_variant_arguments(list_parser)
+        self.add_release_variant_arguments(list_parser, exclude_arches=True)
         list_parser.set_defaults(func=self.release_variant_list)
 
         # CRUD: UPDATE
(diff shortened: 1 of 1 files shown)


pdc_client/plugins/release_variant.py
```diff
@@ -52,14 +52,17 @@ def register(self):
...
+    def add_release_variant_arguments(self, parser, required=False, exclude=False):
         required_args = OrderedDict()
         required_args["release"] = {}
         required_args["uid"] = {}
         required_args["id"] = {}
         required_args["name"] = {}
         required_args["type"] = {}
         required_args["arch"] = {"action": "append", "dest": "arches"}
+        if exclude:
+            required_args.pop("arch")
```
hluk: Argument `exclude` is not very clear. Perhaps:

```python
if requires_arg_arch:
    required_args["arch"] = {"action": "append", "dest": "arches"}
```

hluk (changes requested):
paramter -> parameter

pdc_client/plugins/release_variant.py
xiangge: I was thinking about maybe some parameters also need to pop in the future, so just use the exclude here.
Anyway, use exclude_arches to instead now.

hluk (approved):

simozhan:
LGTM

PR #110, opened 2017-07-31
Increase `pdc content-deliver-repo list` verbosity.

hluk:
JIRA: PDC-1952

### DIFF
--- a/pdc_client/plugins/repo.py
+++ b/pdc_client/plugins/repo.py
@@ -129,7 +129,7 @@ class RepoPlugin(PDCClientPlugin):
         result["Repo Family"] = record["repo_family"]
         result["Content Format"] = record["content_format"]
         result["Content Category"] = record["content_category"]
-        result["Shadow"] = record["shadow"] and "yes" or "no"
+        result["Shadow"] = record["shadow"]
         result["Product ID"] = record["product_id"] or ""
         return result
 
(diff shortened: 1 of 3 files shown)


simozhan:
It seems previous shadow is true/false, but now changed to yes/no, probably the test case failed for this.

hluk:
Hmm, few things:
1. Is the order of columns OK?
2. The size of output table is too big.
3. Column headers are not aligned with the data.
image

simozhan:
@hluk i think the order is not quite important and actually i think maybe ID is not very useful for the user. The size should be very long because most of the repos' name are long, not have any good idea for that.
Maybe name in a new line. Maybe we should discuss it with the users.

xiangge:
2. The size of output table is too big.
Actually for this part, I am working on an issue about the paging, we can keep put all the data as default.
And we can give an option about the paging. 
Will commit patch for reviewing once finish.

hluk:
@xiangge I meant that the table is too wide.

I just added commit that sets minimal column width. It might be a better idea to some library for formatting tables.

xiangge:
@hluk ok, got that.

hluk:
@xiangge @simozhan OK to merge this?

simozhan:
:+1:

simozhan (approved):

xiangge (approved):
ok

PR #111, opened 2017-08-01
Sort commands in pdc --help

simozhan:
JIRA: PDC-2059

### DIFF
--- a/pdc_client/compat.py
+++ b/pdc_client/compat.py
@@ -27,3 +27,38 @@ else:
         return d.iteritems(**kw)
 
     string_type = basestring
+
+
+# copied from py2.7 functools module
+def _total_ordering(cls):
+    """Class decorator that fills in missing ordering methods"""
+    convert = {
+        '__lt__': [('__gt__', lambda self, other: not (self < other or self == other)),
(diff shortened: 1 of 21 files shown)


hluk:
Tests seems to fail with Python 3.x because [relative imports are no longer allowed](https://wiki.python.org/moin/PortingPythonToPy3k#relative-imports). So instead of `import compat` it should be `from pdc_client import compat`.

hluk:
Actually, that's problem when I run tests with Python 3.6. Travis tests give different error.

hluk:
OK, nevermind. Using `from pdc_client import compat` seems to fix the error for Python 3.4 too. Though the error is very cryptic. :)

simozhan:
@hluk cool. thank you very much :)

xiangge:
Looks good to me.

hluk (changes requested):
I would prefer if you leave the original commit untouched -- mainly the author should be the same. I.e. cherry-pick/rebase the original commit and add another commits for the fixes.

simozhan:
@hluk ok. fix it.

xiangge (approved):

hluk (approved):

PR #112, opened 2017-08-03
Fix passing ordering parameter

hluk:
On server-side, Django REST framework uses only the last ordering
parameter, where multiple fields can be separated with comma.


xiangge:
Looks good to me.

xiangge (approved):

simozhan (approved):

PR #114, opened 2017-08-07
Fix content-delivery-repo list ordering

hluk:


xiangge:
Looks good to me. 
One question here why don't create a string with _ORDERING ? 
I think then the join operation can remove.

simozhan (approved):
LGTM :+1:

hluk:
@xiangge I think using `str.join` on list is the cleanest way to get a very long string. Another way is to use [String literal concatenation](https://docs.python.org/2.0/ref/string-catenation.html).

```python
_ORDERING = (
    'variant_arch__variant__release__release_id,'
    'variant_arch__variant__variant_uid,'
    'variant_arch__arch__name,'
    'service,'
    'repo_family,'
    'content_format,'
    'content_category,'
    'shadow,'
    'name'
)
```

But having commas inside strings like this seems bit strange and can be easily missed in IDE.

What do you think?

xiangge:
@hluk  OK, thanks for the explanation, learned.

xiangge (approved):

PR #115, opened 2017-08-08
Add new argument for getting all data

xiangge:
Add new argument "--all" for getting
all data without pagination.
The output is results part.

JIRA: PDC-1234


bin/pdc_client
```diff
@@ -180,5 +182,26 @@ if __name__ == "__main__":
     else:
         with RequestMethod(client, options.resource,
                            options.traceback, options.debug) as request:
-            print(json.dumps(request.dispatch(options.request, data),
+            if options.all:
+                print("You choose get all the data!")
```
hluk: Remove.

hluk (changes requested):
The issue PDC-1234 is about changing PDCClient class, not changing `pdc_client`.

BTW, it's possible to use `page_size=-1` in requests to disable pagination.

```bash
pdc_client -s local -r arches -d '{"page_size":-1}'
```

The drawback is that server has to send all the available data and interrupting client won't change this fact (e.g. interrupting pipe or sending SIGTERM, SIGINT...).

So it's still useful the request one page, print items, request next page etc.

xiangge:
@hluk Thank you for telling me the page_size=-1 can show all the data, I was not aware of that before;
And I agree with you that return all the data might  not be a good way, because it might  return a huge data in some apis. And we also should give an obviously option to let user know how to get all data because they shouldn't know the "page_size=-1" like me. 
And also I think the original behavior should keep, just add new option to let it return all data but not just check if there is a next in the response. 

I think according to the logic of this issue we don't need to change PDCClient codes, just change pdc_client codes can fix this.

What's your thinking?

> The logic would be:
> 1) perform the request
> 2) if response data is not a dict or keys are not previous, next, count and result, return response data directly
> 3) otherwise act as a generator and yield all items from all pages
>

bin/pdc_client
xiangge: I just want to give a comment about showing all the data.
But you are right, should just return data, I have removed it.

hluk:
@xiangge To be clear, the issue is about using the pdc-client as library (which is more useful than calling `pdc_client` when writing Python code).

Example:

```python
client = PDCClient(server)
# yield items (this internally requests next page if needed)
generator = client['arches']._()
for result in generator:
    # result is {"name": "x86_64"} ...
```

Also see the example on JIRA.

xiangge:
@hluk I don't quite understand your mean, in pdc_client, it is using PDCClient class, if people want to use pdc_client as library, they can use within this patch, and if they just want use PDCClient to get all the page, use the logic in the patch are also okay, I think we provide the pdc_client as a tool, not just a library.
And also, like the comment said in jira:

> Examples
> 
> client['releases']['rhel-7.0']()
> ^^^ not paginated, returns just a dict
> 
> client['releases']()
> ^^^ paginated response, automatically returns an iterable of all results
> 
> The logic would be:
> 1) perform the request
> 2) if response data is not a dict or keys are not previous, next, count and result, return response data directly
> 3) otherwise act as a generator and yield all items from all pages

I think what I am implementing is the same like this description. Just put the logic in pdc_client command. As the jira title said "PDC Client should handle pagination automatically", using a “--all”
can make pagination automatically, I think my patch make sense. This is the similar like your example
and also similar like the codes in "pdc" which is called the get_paged function in PDCClient, this is about get all pagination data.

xiangge:
@hluk 
Also discuss with Chuang @simozhan  about this issue, I think I misunderstood the requirement of this issue.
Will close this pr, and try to make another patch to improve the PDCClient class.
Thanks for your help to point out the questions.

xiangge:
As there is no need to add this patch, close this pr

PR #116, opened 2017-08-11
Bug fix for ssl_verify in old pdc_client

simozhan:
The old pdc_client should set ssl_verify to None
to make configuration setting work.
This behavior is the same with new pdc.

Bug: PDC-2092


xiangge (approved):
Looks good to me.

hluk (approved):
In commit message:

    Bug: PDC-2092

should probably be:

    JIRA: PDC-2092

simozhan:
@hluk Ah. Yes. Thanks for pointing this out.

PR #117, opened 2017-08-14
Add the page argument on pdc

xiangge:
Add the page argument on pdc command.

JIRA: PDC-2087


simozhan (changes requested):
Seems some flake8 error needs to be fixed first

xiangge:
@simozhan yes, just forgot to run flake8 check in local. Commit again. Thanks.

pdc_client/__init__.py
```diff
@@ -203,10 +205,21 @@ def get_paged(self, res, **kwargs):
...
+        else:
+            page_size = 20
+        if self.page is not None:
+            allinfo = res(page=self.page, **kwargs)
+            all_count = int(allinfo["count"])
+            if all_count % page_size > 0:
+                total = (all_count / page_size) + 1
+            else:
+                total = all_count / page_size
+            print("Total %d pages, here is the %d page, the count is %s." % (total, self.page, all_count))
```
hluk: Remove. This function shouldn't print anything.

pdc_client/__init__.py
```diff
@@ -203,10 +205,21 @@ def get_paged(self, res, **kwargs):
...
+            return allinfo['results']
```
hluk: Perhaps it's better to return `allinfo` because it's still a single page.

pdc_client/__init__.py
```diff
@@ -203,10 +205,21 @@ def get_paged(self, res, **kwargs):
         """
```
hluk, line 275: The docstring is outdated.

hluk (changes requested):

pdc_client/__init__.py
xiangge: In this get_paged function it just can return the results, like the yield codes:
`yield response['results']`
Because after this function, plugins will list the data by some columns in codes with the returned results.

pdc_client/__init__.py
xiangge: I also thought to remove this line before, but if one chooses to list by page, how to know the total pages without this line?

pdc_client/__init__.py
xiangge: I have update the docstring, please help to review that, and thanks for pointing it.

pdc_client/__init__.py
```diff
@@ -187,10 +189,10 @@ def obtain_token(self):
 
     def get_paged(self, res, **kwargs):
         """
-        This call is equivalent to ``res(**kwargs)``, only it retrieves all pages
-        and returns the results joined into a single iterable. The advantage over
-        retrieving everything at once is that the result can be consumed
-        immediately.
+        This call is equivalent to ``res(**kwargs)``, if there is no self.page
```
hluk: `page` should be passed as argument to this function. It's strange to have it common for all requests of one `PDCClient` instance.

hluk (changes requested):

pdc_client/__init__.py
hluk: If user requests JSON format and wants to process it further, the printed plain text needs to be skipped.

This could be printed in the function which prints out the table, though I'm still not sure if it's good idea.

xiangge: Good catch, if JSON mode, should not print this line.

Currently, it is on the first line, I think it's okay, if I put it in the each print table functions, that means the return value of the get_paged should change, and all the plugins will insert a print line, I also don't think that is a good idea. Which is the better way?

For me, think this get_pages function is just used for our codes, so it's okay to write the print line here. 
Because there is another issue to improve this PDCClient about the paging, I think I can give another patch for common use.

I will modify codes and just let this line print when it is not json output. What'[s your thinking?

pdc_client/__init__.py
simozhan: One thing, to fix PDC-1234, finally we still need to move page from get_paged to init, right?

pdc_client/__init__.py
simozhan: Actually I prefer to delete this print, i am wondering maybe it doesn't matter that users don't know the total page

pdc_client/__init__.py
hluk: @simozhan I don't think that's needed. Following code doesn't make sense from my perspective.

```python
client = PDCClient(server, page=some_page)
```

pdc_client/__init__.py
hluk: OK, I understand, keep it as it is then.

pdc_client/__init__.py
simozhan: @hluk I was assuming in PDC-1234, we will delete get_paged method and merge its codes into __init__, or maybe we will keep get_paged, if that is the case, no problem then

xiangge: @hluk and @simozhan I think we can keep this page parameter in the init function in the future.
But when we call PDCClient in new codes, we don't use it. It is just for old codes. 
By default we just consider people use PDCClient to get all data. So below codes make sence to me.
```
class PDCClient(object):
     def __init__(self, server,  page_size=None):

client = PDCClient(server)
```

pdc_client/__init__.py
xiangge: @simozhan  Yes, based on your and @hluk's advise, I think we can remove this print, if they use page option, they just want to have a look, so it's make sense to me to remove this print line.

xiangge:
@hluk and @simozhan remove the print line, please review that.

pdc_client/__init__.py
```diff
@@ -97,7 +98,7 @@ class PDCClient(object):
     connections. The authentication token is automatically retrieved (if
     needed).
     """
-    def __init__(self, server, token=None, develop=None, ssl_verify=None, page_size=None):
+    def __init__(self, server, token=None, develop=None, ssl_verify=None, page_size=None, page=None):
```
hluk: Please remove the `page` parameter from constructor. It really doesn't make sense for all requests of a PDCClient instance to fetch only particular page.

Correct me if I'm wrong but the `page` attribute is only useful when iterating over results, I don't see any other real use case.

hluk (changes requested):

pdc_client/__init__.py
xiangge: Firstly I want to explain why I want to add this patch. 
This patch is just for pdc command, not for other use, you know in pdc command when you 
use "pdc -s <server> <plugin> list" or "pdc -s <server> --page_size <page_size> <plugin> list" it will return all the data by get_paged.  If there are thousands of data here and I just want to have a look, it is not good to see even it return by pages, it is a long page. So I add this --page <page> to let people can see by page with pdc command. There is another use is about that if people use pdc command in their codes but not PDCClient class and they just want to get by page, this can be used. So from this view, I think the --page should be added in.

So my question is, do you think there is a need to add this patch? If you think the patch needs add, please see why I add patch like this.

> Correct me if I'm wrong but the page attribute is only useful when iterating over results, I don't see any other real use case.

Yes, you are right, the `page` is just used by the get_paged function, from the previous codes you can see that there is  a `page_size` parameter here. This `page` is similar like `page_size`, they are all just used in get_paged function. 
From my view, the get_paged function is just used for our internal command pdc not for other use. And page and page_size in init function are just used for get_paged.  Add parameters is set here is caused by  the current arch of plugins in pdc command, in this command it set page_size first and then add other data in the plugin, so the page just added as this like page_size.
I think they add page_size here but not in plugins, it is because the page_size is a common param, the params in plugins are their own special params. So page_size here, and page also need to add here as get_paged will use, I think this is make sense.

As we have already added a new class to support multi pages, so if others use the PDCClient, they just use PDCClient's call function to list all data or add page page_size or other things in kwargs. Like:
` 
client = PDCClient(server)
client["releases"](**kwargs)
`
You can see user don't need to add the page and page_size when they init a new object.
And they just need to add the page and page_size in kwargs if they want.

According to this, if you still think the page should not add here, I think so the page_size is, and this should be another issue, we should make other pr or patches for this, but not change it in this patch.

hluk: OK, now I understand.

But we shouldn't add private/internal stuff to public API. `PDCClient.__init__()` is part of public API.

Having `page_size` there makes sense since it can be used for all request. But it's only used in `get_paged()` and not in `__getitem__()` (I didn't notice this before). This looks like a bug to me.

Even though the parameter is not mentioned in docstring, which it should be, user can make quick assumptions about it based on its name. We should make it harder for the API to be misused.

For now, it would be better to set the parameter to `PDCClient` using some private interface. But I'm not sure what's the best way to solve this in Python.

xiangge: > Having page_size there makes sense since it can be used for all request. But it's only used in get_paged() and not in __getitem__() (I didn't notice this before). This looks like a bug to me.

Yes, currently the page_size is just used by get_paged(), but  the page_size currently can set in plugins instead, for example if you set the page_size in plugin's get list function.
 Like ProductPlugin.product_list, it will get the page_size here.
```
   def product_list(self, args):
        filters = {}
        if args.inactive:
            filters['active'] = False
        elif not args.all:
            filters['active'] = True
        filters['page_size'] = True

        products = self.client.get_paged(self.client["products"]._, **filters)
```

So I said the page_size here is same like page.

And with another patch I committed, it should be:
```
client = pdc_client.PDCClient("upstream", is_multi=True)

a = client["releases"](page_size=30)
```

xiangge: Add new class here for the page option of pdc command.
Please help to review.

pdc_client/__init__.py
```diff
@@ -185,12 +186,45 @@ def obtain_token(self):
...
+
+        Once you set the comment, it will be sent in all subsequent requests.
+
+        :param comment:     what comment to send to the server
+        :paramtype comment: string
+        """
+        self.session.headers["PDC-Change-Comment"] = comment
+
+
+class PDCClientWithPage(PDCClient):
```
hluk: Can you hide the new class? E.g. move it to runner.py. I don't think it's useful to anyone but the `pdc` command.

pdc_client/__init__.py
```diff
@@ -185,12 +186,45 @@ def obtain_token(self):
...
+    def __init__(self, server, token=None, develop=None, ssl_verify=None, page_size=None, page=None):
+        """
+        Create new client instance with page prarameter.
+        Other params are all used for base class.
+        :param page:  the page number of the data.
+        """
+        super(PDCClientWithPage, self).__init__(server, token, develop, ssl_verify, page_size)
+        self.page = page
+
     def get_paged(self, res, **kwargs):
```
hluk: Removing a public method from `PDCClient` means that next API version should be at least 1.8.x.

But I think it's better to keep it there and just mark it as deprecated for now (and point to a new interface).

pdc_client/test_helpers.py
```diff
@@ -44,7 +44,7 @@ def __getitem__(self, key):
     class MockAPI(object):
         """Representation of mocked API.
 
-        This class replaces the PDCClient in code. Without any additional
+        This class replaces the PDCClientWithPage in code. Without any additional
```
hluk: I would rather see separate small test for `PDCClientWithPage` (this class should be private anyway).

hluk (changes requested):

pdc_client/test_helpers.py
xiangge: All the test cases are using PDCClientWithPage now, because the pdc use a runner.py, and runner.py calls PDCClientWithPage.

pdc_client/__init__.py
xiangge: Do you mean like these codes in class PDCClient?
```
def get_paged(self, res, **kwargs):
    # obsolete
    pass
```

xiangge: Rollback the get_paged codes in PDCClient, and add new codes with self.page in new PDCClientWithPage. Please review

pdc_client/test_helpers.py
hluk: The docstring says that `MockAPI` is "Representation of mocked API" which would indicate `PDCClient`. The class `PDCClientWithPage` is just helper for `pdc` command.

xiangge: Yes, correct, I have changed that, please help to review.

hluk (approved):

simozhan (approved):

simozhan:
LGTM

xiangge:
As patch is committed, close this pr, not auto close maybe because use --amend to commit codes and use --force to push codes in this branch.

PR #118, opened 2017-08-15
Update spec file from downstream

hluk:


pdc-client.spec
```diff
@@ -1,22 +1,35 @@
...
-%global with_python3 1
+# If the definition isn't available for python3_pkgversion, define it
+%{?!python3_pkgversion:%global python3_pkgversion 3}
+%bcond_without  python3
 %else
+%bcond_with     python3
+%endif
+
+# Compatibility with RHEL. These macros have been added to EPEL but
+# not yet to RHEL proper.
```
simozhan: FYI. When we make pdc related codes upstream, we decide to not mention any redhat related info in comments/codes/tests.

pdc-client.spec
```diff
@@ -1,22 +1,35 @@
...
+%{!?python2_sitelib: %global python2_sitelib %(%{__python2} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")}
+%{!?python2_sitearch: %global python2_sitearch %(%{__python2} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib(1))")}
 %{!?py2_build: %global py2_build %{expand: CFLAGS="%{optflags}" %{__python2} setup.py %{?py_setup_args} build --executable="%{__python2} -s"}}
 %{!?py2_install: %global py2_install %{expand: CFLAGS="%{optflags}" %{__python2} setup.py %{?py_setup_args} install -O1 --skip-build --root %{buildroot}}}
-%endif
 
 Name:           pdc-client
 Version:        1.2.0
-Release:        1%{?dist}
+Release:        4%{?dist}
```
simozhan: I remember Jan mentioned that our version should be 1.7.0-1, any consideration for this version?

hluk: @simozhan This is just update from dist-git. I'll post another PR for 1.7.0-1.

simozhan:
@hluk I am still not quite understand about this patch, can you describe a bit more with this patch, why we have this patch, and why not post PR 1.7.0-1 directly? Thanks

xiangge:
What is the dist-git link you mentioned?
And currently, the `git tag` just shows the latest 1.2.0-1, there is not -2, -3 -4 version here.

hluk:
@simozhan I want to have a separate commit for this since these are not my changes and the repo should already contain up-to-date spec file when moving to next version.

> What is the dist-git link you mentioned?

@xiangge Sorry, not dist-git, [pdc-client in Pagure](https://src.fedoraproject.org/rpms/pdc-client/blob/master/f/pdc-client.spec).

> And currently, the git tag just shows the latest 1.2.0-1, there is not -2, -3 -4 version here.

Yes, those versions are only available in dist-git (`rhpkg clone pdc-client`).

pdc-client.spec
hluk: OK, I'll remove the comments.

pdc-client.spec
simozhan: Ok. I got the point. No problem then

simozhan (approved):
Finally I got the point. This downstream is the code in pagure. I was assuming it is redhat internal codes. Sorry about that. As this is only for spec file change from downstream, It looks good to me.

simozhan:
The new added patches are also good to me.

hluk:
I've updated the version to 1.7.0-1 in this commit.

I also noticed that the tag automatically created by tito is `pdc-client-1.7.0-1`. I would like to do the same in internal repo (since it already contains tag `1.7.0-1` for the server part). Do you thinks it's OK?

xiangge (approved):

xiangge:
Looks good to me.

xiangge:
Add for the internal changes, I think it is ok.

simozhan:
@hluk Ok for me

PR #119, opened 2017-08-21
Add MultiPageBeanBag class to support get multi pages

xiangge:
Add MultiPageBeanBag class to support get multi pages

JIRA-1234

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -14,9 +14,9 @@ from os.path import expanduser, isfile, isdir
 import sys
 from subprocess import Popen, PIPE
 
-import beanbag
 import requests
 import requests_kerberos
+from beanbag import BeanBag, BeanBagException
 
 GLOBAL_CONFIG_DIR = '/etc/pdc.d/'
 USER_SPECIFIC_CONFIG_FILE = expanduser('~/.config/pdc/client_config.json')
@@ -90,6 +90,29 @@ def read_config_file(server_alias):
     return result
 
 
+class NoResultsError(Exception):
+    """ Exception for getting all pages of data
+        Raise this NoResultsError if there is an unexpected data
+        returned when get all pages of data by results() function.
+
+        Data members:
+             * response -- response object
+    """
+    def __init__(self, response):
+        """Create a NoResultsError"""
+
+        self.response = response
+
+    def __repr__(self):
+        return "%s(%r)" % (self.__class__.__name__, self.response)
+
+    def __str__(self):
+        msg = "Unexpected data here"
+        if self.response:
+            msg = "%s - response: %s" % (msg, str(self.response))
+        return msg
+
+
 class PDCClient(object):
     """BeanBag wrapper specialized for PDC access.
 
@@ -97,8 +120,25 @@ class PDCClient(object):
     interface that can use configuration files for specifying server
     connections. The authentication token is automatically retrieved (if
     needed).
-    """
+    ::
+            # Example: Get per page of release
+            client = PDCClient(<server>)
+            client.releases._()
+            client["releases"]._()
+            ...
 
+            # Example: create one release data
+            client["releases"]._(<dict data>)
+            ...
+
+            # Example: Iterate all pages of releases
+            # Will raise NoResultsError with response when return unexpected result.
+            try:
+                for r in client["releases"].results():
+                    ...
+            except NoResultsError as e:
+                # handle e.response ...
+    """
     def __init__(self, server, token=None, develop=None, ssl_verify=None, page_size=None):
         """
         Create new pdc client instance.
@@ -168,7 +208,7 @@ class PDCClient(object):
 
         content_type = "application/json"
         encode = json.dumps
-        self.client = _BeanBagWrapper(beanbag.BeanBag(url, session=self.session, fmt=(content_type, encode, decode)),
+        self.client = _BeanBagWrapper(BeanBag(url, session=self.session, fmt=(content_type, encode, decode)),
                                       page_size)
         if not develop:
             # For develop environment, we don't need to require a token
@@ -188,7 +228,7 @@ class PDCClient(object):
         for end_point in token_end_points:
             try:
                 return self.auth[end_point]._()['token']
-            except beanbag.BeanBagException as e:
+            except BeanBagException as e:
                 if e.response.status_code != 404:
                     raise
         raise Exception('Could not obtain token from any known URL.')
@@ -201,9 +241,7 @@ class PDCClient(object):
 
         :param res:     what resource to connect to
         :param kwargs:  filters to be used
-
         ::
-
             # Example: Iterate over all active releases
             for release in client.get_paged(client['releases']._, active=True):
                 ...
@@ -254,6 +292,7 @@ class _BeanBagWrapper(object):
        This class wraps attributes and items of Beanbag to let page_size of
        PDCClient's constructor work.
     """
+
     def __init__(self, client, page_size):
         self.client = client
         self.page_size = page_size
@@ -268,6 +307,29 @@ class _BeanBagWrapper(object):
 
     def __getitem__(self, *args, **kwargs):
         return _BeanBagWrapper(self.client.__getitem__(*args, **kwargs), self.page_size)
+
+    def results(self, *args, **kwargs):
+        """
+           Return an iterator with all pages of data.
+           Return NoResultsError with response if there is unexpected data.
+        """
+        def worker():
+            kwargs['page'] = 1
+            while True:
+                response = self.client(*args, **kwargs)
+                if isinstance(response, list):
+                    yield response
+                else:
+                    yield response['results']
+                if isinstance(response, dict):
+                    if set(response.keys()) != set(['count', 'previous', 'results', 'next']):
+                        raise NoResultsError(response)
+                    if response['next']:
+                        kwargs['page'] += 1
+                else:
+                    break
+
+        return itertools.chain.from_iterable(worker())
 
 
 class PDCClientWithPage(PDCClient):


pdc_client/__init__.py
```diff
@@ -89,8 +94,63 @@ def read_config_file(server_alias):
...
+
+            elif len(args) == 1:
+                verb, body = "POST", args[0]
+            elif len(args) == 2:
+                verb, body = args
+            else:
+                raise TypeError("__call__ expected up to 2 arguments, got %d" % (len(args)))
+            return self.make_request(path, verb, kwargs, body)
+        except Exception:
+            raise
```
hluk: Remove unnecessary try-except.

pdc_client/__init__.py
```diff
@@ -89,8 +94,63 @@ def read_config_file(server_alias):
     return result
 
 
+class MultiPageBeanBag(BeanBag):
```
hluk: Can you add a docstring here?

pdc_client/__init__.py
```diff
@@ -89,8 +94,63 @@ def read_config_file(server_alias):
...
+                result = self.make_request(path, verb, kwargs, body)
+
+                if self.is_multi_page_response(result) and ("page" not in kwargs.keys()):
+                    count = result["count"]
+                    multi_result = result['results']
+                    while result['next']:
+                        next_page = self.url2path_ext(result['next'])
+                        kwargs["page"] = next_page["page"][0]
+                        result = self.make_request(path, verb, kwargs, body)
+                        multi_result.extend(result['results'])
```
hluk: Use generator instead (as in previous code). With `yield`, user can get the first results more quickly with smaller `page_size`.

pdc_client/__init__.py
```diff
@@ -180,44 +240,26 @@ def obtain_token(self):
...
-            # Example: Iterate over all active releases
-            for release in client.get_paged(client['releases']._, active=True):
-                ...
         """
         if self.page_size is not None:
             kwargs['page_size'] = self.page_size
-
             if self.page_size <= 0:
                 # If page_size <= 0, pagination will be disable.
                 return res(**kwargs)
```
hluk: You can remove this `if` branch.

hluk (changes requested):

pdc_client/__init__.py
xiangge: I think here we use the yield or while both ok, because even use yield we will final return one json string but not multi json string, so yield should extend all jsons then return.
And the original patch use while.

pdc_client/__init__.py
xiangge: Good idea, I will remove it.

pdc_client/__init__.py
xiangge: After discuss with @simozhan , I will make the change in make_request function as the original patch.

pdc_client/__init__.py
xiangge: OK, will add one

pdc_client/__init__.py
simozhan: Agree with hluk

pdc_client/__init__.py
simozhan: Yeah. Better to add docstring here.

pdc_client/__init__.py
```diff
@@ -89,8 +94,63 @@ def read_config_file(server_alias):
...
+
+    def call(self, path, *args, **kwargs):
+        """
+        Re-write the call function of BeanBag.
+        Let it default get multi pages data if there is no page parameter.
+        """
+        try:
+            if len(args) == 0:
+                verb, body = "GET", None
+                result = self.make_request(path, verb, kwargs, body)
```
simozhan: Can we move these codes out and deal with multi page response in make request like following codes did. I think maybe that is neat. https://github.com/lkocman/beanbag/commit/407fff011c9e1f3883734f44a3d48fe041b23239#diff-5ad74912650ff9cf4a6754f8176f68faR336

pdc_client/__init__.py
xiangge: OK, after look into the yield codes in get_pages, I think this have a better user experience when user uses pdc list command with the formatted output. I will change code with yield and commit then.

xiangge:
Change patch, please review.

pdc_client/__init__.py
```diff
@@ -161,7 +258,13 @@ def decode(req):
 
         content_type = "application/json"
         encode = json.dumps
-        self.client = beanbag.BeanBag(url, session=self.session, fmt=(content_type, encode, decode))
+        # In get_paged function it returns an interator, so user can get each pages quickly in plugins.
+        # The original call function of BeanBag returns the list.
+        # Keep the current codes and options in codes. Just init MultiPageBeanBag when is_multi is true.
+        if is_multi:
+            self.client = MultiPageBeanBag(url, session=self.session, fmt=(content_type, encode, decode))
```
hluk: Why not use the new class without having the extra option and have `yield` there?

hluk (changes requested):

pdc_client/__init__.py
xiangge: Because if we use the new class here, the worker() in get_paged function should be removed as the new class can get all the pages.  

And as I comment on codes, the call function return a list, when you use pdc command to list data, the list function in plugins will get an interator return from get_pages.

As you mentioned, you prefer to let user quickly get first page when they list multi page. The worker() can do this thing but it can't add into the new class as there is a different return value.

So keep the original get_paged codes to let it return quickly when use pdc to get data by column but not json. 

And also I think we just provide a base class for users to use in there codes, and for our functions, we don't need to make change, because as you know for pdc and pdc_client  they can get all pages. 
And we have no need to modified the __init__ codes here, and a make_request is used by the call function and we also have no need to modify that function I think this is a more reasonable  to modify the codes, but not add an extra option in new class.

hluk: > Because if we use the new class here, the worker() in get_paged function should be removed as the new class can get all the pages.

The new class gets all pages but client needs to wait until all of them are fetched regardless of the `page_size` field value (with small `page_size` you would only make more request but client still needs to wait for all pages to be fetched). I think it could be better to use `page_size=-1` so PDC server can send all results at once. But I still don't like that the `page_size` is ignored.

> And as I comment on codes, the call function return a list, when you use pdc command to list data, the list function in plugins will get an interator return from get_pages.

If it's not possible to return different type (other than list) in the the new class, is it possible to handle this in in `PDCClient.__getitem__()`?

BTW, I cannot find `BeanBag.make_request()` in [documentation](http://beanbag.readthedocs.io/en/latest/v1.html) and overriding the class methods is not mentioned at all. It could make upgrading to a new version difficult.

xiangge: > The new class gets all pages but client needs to wait until all of them are fetched regardless of the page_size field value (with small page_size you would only make more request but client still needs to wait for all pages to be fetched). I think it could be better to use page_size=-1 so PDC server can send all results at once. But I still don't like that the page_size is ignored.

Yes, I think -1 also works for this case, but I think if we use page_size=-1, the data will still return after long time, because server should query all the data first. Maybe this is a little faster than than sending several requests.
But when the data is very big, maybe the http server has a time out and then return nothing, so I think return the whole data in one request maybe not a better way than the current changes, what's your thinking?

xiangge: > If it's not possible to return different type (other than list) in the the new class, is it possible to handle this in in PDCClient.__getitem__()?

I think we can't choose the type in __getitem__, because in this function it returns the value of instance's key,  the type we are talking is about the return value of BeanBag.call, do you have any ideas about this part? 

> BTW, I cannot find BeanBag.make_request() in documentation and overriding the class methods is not mentioned at all. It could make upgrading to a new version difficult.

The make_request function is instance method, so you need init a instance first, and also this is used as a return of  call function in BeanBag.
```
>>> import beanbag
>>> beanbag.__version__
'1.9.2'
>>> beanbag.BeanBag
<class 'beanbag.url_v1.BeanBag'>
>>> beanbag.BeanBag.make_reuqest
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: type object 'BeanBag' has no attribute 'make_reuqest'
>>> a = beanbag.BeanBag
>>> a = beanbag.BeanBag()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: <lambda>() takes at least 2 arguments (1 given)
>>> a = beanbag.BeanBag("aaa","nn")
>>> a.make_request
<BeanBagBase(aaa/make_requestnn)>
>>> 
```

xiangge: @hluk Is that ok to merge? or is there anything which still  need change from your point of view?

xiangge: @hluk any updates?

hluk: Following code seems extremely slow (for me it takes ~2 seconds to fetch single page).

```python
import pdc_client
c = pdc_client.PDCClient('qa', is_multi=True)
c['releases']()
```

I.e. doing multiple requests could take forever. I suggest overriding `page_size=-1` ... but that would actually mean that the new class is not needed.

We should probably talk about this tomorrow over phone/video.

pdc_client/__init__.py
```diff
@@ -109,6 +204,8 @@ class to perform requests.
         :param ssl_verify: True for validating SSL certificates with system CA
                            store; False for no validation; path to CA file or
                            directory to use for validation otherwise
+        :param is_multi:   This is about if the new object uses the MultiPageBeanBag
```
hluk: Users of this API don't need to know about `MultiPageBeanBag` -- the class should be private and hidden from users.

More descriptive argument name would be nice, something like `get_all_pages` or `paged` (default `True`).

hluk (changes requested):

pdc_client/__init__.py
xiangge: I think it is ok to add the public class here, then user can choose the way to create codes they like and also this doesn't break the previous codes and logic.

Also I think is_multi is more directly  than the get_all_pages and paged, as this is a boolean

pdc_client/__init__.py
xiangge: Based on this issue, the requirement is about  the logic like this. So this class is still needed.
And also the test in your case is just for less data. If for large data, get data once a time maybe have issue about the http request.
Also even in this case you can also try like this:

```
import pdc_client
c = pdc_client.PDCClient('qa')
c['releases'](page_size="-1")
```
Let's talk about this by phone/video then.

pdc_client/__init__.py
hluk: > I think it is ok to add the public class here

Don't add stuff to public API unless really necessary. We need to make sure that changing the public API won't break existing code (otherwise users need to update them). And as I mentioned above `BeanBag.make_request()` is undocumented (not part of BeanBag's API?) so how do we know that we are re-implementing it correctly.

> Also I think is_multi is more directly than the get_all_pages and paged, as this is a boolean

Sorry but I cannot tell what `is_multi` means just by looking at the name.

xiangge: > Don't add stuff to public API unless really necessary. We need to make sure that changing the public API won't break existing code (otherwise users need to update them). And as I mentioned above BeanBag.make_request() is undocumented (not part of BeanBag's API?) so how do we know that we are re-implementing it correctly.

For this part I have made sure that this can't break the codes, and the make_request is the correct api of BeanBag, I will show you then in latter video/talk.

> Sorry but I cannot tell what is_multi means just by looking at the name.
If so I think, is_multi_page should be better, what's your thinking? and also this argument is described in docstring.

xiangge:
@lubomir Hi, lubomir, do you have time to help review this patch. Because I saw you file this issue to let pdc-client add this patch in.
For the original codes, we can provide a way to get all  the pages of data. Like:
```
import pdc_client
c = pdc_client.PDCClient('qa')
c['releases'](page_size="-1")
```
Why we still need to add such patch to let pdc-client send multi requests to get data in while loop?

pdc_client/__init__.py
```diff
@@ -89,15 +95,104 @@ def read_config_file(server_alias):
...
 
+class MultiPageBeanBag(BeanBag):
+    """
+    Add this class to support to return all pages of data by default in a get request.
+
+    This class is based on the patch:
+    https://github.com/lkocman/beanbag/commit/407fff011c9e1f3883734f44a3d48fe041b2323
+    9#diff-5ad74912650ff9cf4a6754f8176f68faR336
+
+    We can use the class directly if the patch adds in the master of beanbag in the futrue.
```
lubomir: I would just remove this line. There's a typo: `futrue` → `future`.

pdc_client/__init__.py
```diff
@@ -161,7 +258,13 @@ def decode(req):
...
         content_type = "application/json"
         encode = json.dumps
-        self.client = beanbag.BeanBag(url, session=self.session, fmt=(content_type, encode, decode))
+        # In get_paged function it returns an interator, so user can get each pages quickly in plugins.
+        # The original call function of BeanBag returns the list.
+        # Keep the current codes and options in codes. Just init MultiPageBeanBag when is_all_page is true.
+        if is_all_page:
+            self.client = MultiPageBeanBag(url, session=self.session, fmt=(content_type, encode, decode))
+        else:
+            self.client = BeanBag(url, session=self.session, fmt=(content_type, encode, decode))
```
lubomir: It would be nice to emit some deprecation warning for the BeanBag path, so that at some point it can be removed.

pdc_client/__init__.py
```diff
@@ -89,15 +95,104 @@ def read_config_file(server_alias):
...
+        returns path, params
+        """
+        parsed = urlparse(url)
+        return parse_qs(parsed.query)
+
+    def _make_request(self, path, verb, params, body):
+        """
+        Rename the make_request of BeanBag to _make_request.
+        As the class type is MultiPageBeanBagBase, we can't
+        use a super function.
```
lubomir: I'm not sure I understand this comment. `MultiPageBeanBagBase` is a subclass of `BeanBag`, so using `super` would call the method on parent, right?

xiangge: In BeanBag class if you create a subclass, the base class will change to <classname>+Base, I saw this is inside beanbag __new__ function. And a super function will return error as the base class is not BeanBag.
So we just rename the make_request like what original patch does, not call super.make_request in the new function.
That's the comment's meaning.

pdc_client/__init__.py
xiangge: Actually I think we just use these two branch, if user wants to get all data, just set is_all_page to true and use MultiPageBeanBag, otherwise just use the original codes to BeanBag

pdc_client/__init__.py
xiangge: Changed, please review it.

lubomir (changes requested):
We discussed this with @hluk and came up with a better design for this.

Instead of adding a parameter to the constructor, it would be cleaner to create a new subclass of `PDCClient` that handles pagination automatically. It would basically use the existing `get_paged` method inside its `__call__` implementation.

Let's demonstrate on an example with listing active releases:

```python
# Current status
c = PDCClient()
# returns the first page including the count and prev/next links
c.releases._(active=True)
# Iterating over all releases page after page:
for release in c.get_paged(c['releases']._, active=True):
    pass

# Proposed solution
c = PagingPDCClient()
for release in c.releases._(active=True):
    pass
```

The advantage of this is that it preserves the iterator and the results can be consumed without waiting for all the data to be downloaded. Backwards compatibility is preserved.

The semantics of the new class would be like this:
 * for `GET` requests that look like a paged result it would work as an iterator yielding parts of the response (elements of the `results` key)
 * for other methods than `GET` or non-paged `GET` responses it would behave identically to regular `PDCClient` class

What do you think?

xiangge:
@lubomir and @hluk Thanks for your discussion and response.

> Instead of adding a parameter to the constructor, it would be cleaner to create a new subclass of PDCClient that handles pagination automatically. It would basically use the existing get_paged method inside its __call__ implementation.

What do you mean of a subclass?  
We should keep the original codes get_paged and BeanBag part for the internal pdc command using, right?
Do you mean keep two class here, one is PDCClient for return single page and the other is PagingPDCClient?
If so, do you think it's better that we just keep one PDCClient, and set a parameter here to let ct chose if he wants a paging.  Current codes, I just use a parameter to sign this. It makes codes simple to look and use.

```
# Current status
c = PDCClient()
# returns the first page including the count and prev/next links
c.releases._(active=True)
# Iterating over all releases page after page:
for release in c.get_paged(c['releases']._, active=True):
    pass

c = PDCClient(is_all_page=True)
for release in c.releases._(active=True):
    pass
```

For implementing get_paged in __call__ function part, I have tried that before, because as you know, in beanbag the call function return a list data, and the get_paged function returns an iterator data, and the reason why you can see the data return by pages with a pdc list command is that when you don't choose a json format return in plugin codes, it will return with enumerate, then you can see the data returns one page by one page, but if you choose return json, you can see the final data is a list which likes the current beanbag returns, so in this situation the data will return one time after the yield till the end. 

Here is the codes in plugins/product.py
```
   if args.json:
            print(self.to_json(list(products)))
            return

        fmt = '{0:25} {1:35} {2}'
        for num, product in enumerate(products):
            data = self.prep_for_print(product)
            if num == 0:
                print(fmt.format(*data.keys()))
            print(fmt.format(*data.values()))

```

And for my point of view, I think when ct choose to use with the PDCClient class directly, they just want to write it in codes, and they want get a json data, that mean they don't care if it returns one page by one page or returns once a time. The final result is to get all data.

And also we can implement the yield part in MultiPageBeanBag, actually I have already tried that before, if we write the codes like get_paged in the make_request function of  MultiPageBeanBag, the __call__ function can return the same data like get_paged. That means ct will use a enumerate before they loop the return data, because seems it returns errors when we just use `for release in c.releases._(active=True)`

```
# Proposed solution
c = PagingPDCClient()
for release in enumerate(c.releases._(active=True)):
    pass
```
or

```
c = PDCClient(is_all_page=True)
for release in enumerate(c.releases._(active=True)):
    pass
```
I am not sure if it is easy for ct to use. 
Or maybe we can just let the __call__ function return like 
enumerate(get_paged returning data), but not sure if it can change to json data in ct's codes.
Even write like this I still don't sure if it is easy for ct to use.

So I prefer just let the codes return as a list, like the current codes. Then ct can use below codes:
```
c = PDCClient(is_all_pagei=True)
c.releases._(active=True)): # return all data
```

What's your thinking? 
Please point to me if I am wrong,  I am not sure but maybe your ideas more correct than what I am thinking about.

hluk:
Few points:
- Don't override `BeanBag.make_request()` (the method is not documented).
- Class `MultiPageBeanBag` should be private, no need for API user to instantiate it.
- Blocking and making multiple requests in loop takes too much time. Using `page_size=-1` is much faster and also returns list of results directly.
- Instead of using the new argument, sub-classing `PDCClient` looks like cleaner solution because you don't need to change the original class and don't have to use additional `if` statement for testing new argument in constructor.
- I still don't like `is_all_page` name much more than `is_multi` -- perhaps `omit_paging`?

Personally I wouldn't mind using following.

```python
# Optionally set page_size=-1 in constructor to request all results.
c = PDCClient(server)

# PDCClient.results() yields items immediately when page is available;
# throws if the response doesn't contain a list or a dict with results
# i.e. set(response.keys()) == {"results", "next", "previous", "count"}.
for release in c["releases"].results(base_product="rhel-7"):
    # ...
```

It's safer and easier to read since it explicitly expects items to be in response.

This would require to wrap stuff like `c["releases"]` in another `PDCClient` (or a new class that provides the new method) -- which is required anyway to make `page_size=-1` work from constructor (currently broken).

lubomir:
> Blocking and making multiple requests in loop takes too much time. Using page_size=-1 is much faster and also returns list of results directly.

This is a key point. As it is now, the patch does not really solve the problem of making it easy to get all results and acting on them as soon as possible without having to wait for all of them.

I didn't realize it's not so easy to override the behaviour, since `c.releases` is not actually an instance of `PDCClient` anymore, but a `BeanBag` object.

I also like @hluk's proposal for explicit `results` method. That's indeed very explicit. The only downside I can think of is that if at some point there's a URL that ends with `/results`, it would have to be called as `client['something']['results']` and not `client.something.results`. Not a big loss in my opinion.

hluk:
@lubomir It should be easy to wrap both results of `__getitem__()` and `__getattr__()` with `PDCClient` in the same way, i.e. both `client['something']` and `client.something` would work. So no loss at all. :)

BTW, I don't use `client.something` because it leads to inconsistencies in code since you cannot access for example `release-variants` this way (though we could solve this in `__getattr__()` by replacing `_` with `-` in attribute name and passing it to BeanBag's `__getitem__()` instead).

xiangge:
@hluk and @lubomir , commit a new patch based on #128, please help to review.
 Also I didn't implement as the iterator  which likes get_paged, I think this is not easy for ct to use, just let the call() return list as before, what's your thinking of this part?

xiangge:
@hluk, @simozhan base patch committed, please help to review this.

pdc_client/__init__.py
```diff
@@ -237,3 +249,58 @@ def set_comment(self, comment):
...
+
+        result = self.client(*args, **kwargs)
+        if self.omit_paging and self.is_multi_page_response(result) and ("page" not in kwargs.keys()):
+            count = result["count"]
+            multi_result = result['results']
+            while result['next']:
+                next_page = self.url2path_ext(result['next'])
+                kwargs["page"] = next_page["page"][0]
+                result = self.client(*args, **kwargs)
+                multi_result.extend(result['results'])
```
hluk: This seems to work the same way as `page_size=-1` but is slower since this sends multiple requests.

hluk (changes requested):
@xiangge 
> Also I didn't implement as the iterator which likes get_paged, I think this is not easy for ct to use

I don't see how it's not easy to use.

```python
for result in client['x']['y']['z'].results(arg1='X', arg2='Y'):
    # handle results
```

Any concerns?

xiangge:
@hluk I did some tests on qe env.
Here is the result 

>  (pdc) [chcao@dhcp-128-226 pdc-client]$ python tests.py qe
> This with the default 200 items per page:
> This is the start time: 1506675940.13
> This is the end time: 1506677808.73
> Run the func use : 1868.59404016
> 
> This with the default 20 items per page:
> This is the start time: 1506677808.73
> This is the end time: 1506678222.18
> Run the func use : 413.451629162
> 
> This with all items per page:
> This is the start time: 1506678222.18
> This is the end time: 1506680452.35
> Run the func use : 2230.16915798

```
#!/usr/bin/python
import time
import sys
import requests
import pdc_client

from pdc_client import PDCClient

server = sys.argv[1]
requests.packages.urllib3.disable_warnings(
                requests.packages.urllib3.exceptions.InsecureRequestWarning)
client_once_get_all = PDCClient(server, page_size=-1)
client = PDCClient(server, page_size=200, omit_paging=True)
client_with_default = PDCClient(server, omit_paging=True)

def timer(func):
    def decor():
        start_time = time.time();
        print("This is the start time: %s" % start_time)
        func();
        end_time = time.time();
        print("This is the end time: %s" % end_time)
        d_time = end_time - start_time
        print("Run the func use : %s" % d_time)
    return decor

print("This with the default 200 items per page:")
@timer
def t1():
    client["composes"]._()
t1()

print
print("This with the default 20 items per page:")
@timer
def t2():
    client_with_default["composes"]._()
t2()
print
print("This with all items per page:")
@timer
def t3():
    client_once_get_all["composes"]._()
t3()
print
```

xiangge:
@hluk I have commit new patch with generator version, for compare I keep the original list version codes in comment, will remove it if the patch passes the review.
And here is the test result in local env with new codes, please review.

> (pdc) [chcao@dhcp-128-226 pdc-client]$ python tests.py 
> Get data once a time
> This is the start time: 1506757210.87
> This is the end time: 1506757370.17
> Run the func use : 159.296233892
> Get data 200 per time
> This is the start time: 1506757370.17
> This is the end time: 1506757464.95
> Run the func use : 94.7827880383
> Get data by default page_size
> This is the start time: 1506757464.95
> This is the end time: 1506757546.73
> Run the func use : 81.775053978
> 

```
#!/usr/bin/python
import time
import sys
import requests
import pdc_client

from pdc_client import PDCClient

# server = sys.argv[1]
server = "upstream"
#requests.packages.urllib3.disable_warnings(
#                requests.packages.urllib3.exceptions.InsecureRequestWarning)
client_once_get_all = PDCClient(server, page_size=-1)
client = PDCClient(server, page_size=200, omit_paging=True)
client_with_default = PDCClient(server, omit_paging=True)

def timer(func):
    def decor():
        start_time = time.time();
        print("This is the start time: %s" % start_time)
        func();
        end_time = time.time();
        print("This is the end time: %s" % end_time)
        d_time = end_time - start_time
        print("Run the func use : %s" % d_time)
    return decor

print("Get data once a time")
@timer
def t1():
    results = []
    for r in client_once_get_all["composes"]._():
        # handle results
        results.extend(r)
    # print(results)
t1()


print("Get data 200 per time")
@timer
def t2():
    results = []
    for r in client["composes"]._():
        # handle results
        results.extend(r)
    # print(results)
t2()

print("Get data by default page_size")
@timer
def t3():
    results = []
    print(client_with_default["composes"]._())
    for r in client_with_default["composes"]._():
        # handle results
        results.extend(r)
    # print(results)
t3()
```

pdc_client/__init__.py
xiangge: I just think people will write a for loop to get all the data. But currently we just return the whole data.
But as you said return as a iterator saves more space of pdc_client, that makes sense,  have changed the codes

hluk (changes requested):
Rebase the code first and remove the commented-out code.

Not sure if it's OK to have result of `__call__()` to depend on an argument in constructor. In code this could be difficult to read. Having separate method for this seems better to me.

xiangge:
@hluk  OK, as this patch makes sense for you, will remove the comments.

pdc_client/__init__.py
```diff
@@ -254,20 +254,37 @@ class _BeanBagWrapper(object):
...
+    def __init__(self, client, page_size, omit_paging=False):
         self.client = client
         self.page_size = page_size
+        self.omit_paging = omit_paging
 
     def __call__(self, *args, **kwargs):
         if 'page_size' not in kwargs:
             kwargs['page_size'] = self.page_size
-        return self.client(*args, **kwargs)
+        if not self.omit_paging or "page" in kwargs.keys() and self.page_size < 0:
```
hluk: Even though operator `or` has lower precedence than `and` [1] this is not a well known fact and this code is confusing. (Is it **(X or Y) and Z** or **X or (Y and Z)**?)

Use parentheses to make the intent clear.

[1] https://docs.python.org/2/reference/expressions.html#operator-precedence

pdc_client/__init__.py
```diff
@@ -254,20 +254,37 @@ class _BeanBagWrapper(object):
...
+            result = self.client(*args, **kwargs)
+        else:
+            def worker():
+                kwargs['page'] = 1
+                while True:
+                    response = self.client(*args, **kwargs)
+                    yield response['results']
```
hluk: This throws an exception if the response contains different data.

hluk (changes requested):
As I mentioned before, I don't like having to depend on a argument in constructor and to have the call function to return generator conditionally.

It would be much better to have separate function which always returns the same type of result, in this case generator for items. I think this can be implemented as iterator (i.e. `_BeanBagWrapper.__iter__()`)

Also if server sends unexpected response it would be nice for user to be able to handle it, e.g.

```python
if isinstance(response, list):
    # Handle simple unpaged list of results.
# ...otherwise always verify response before iterating.
if response.keys() != {'count', 'next', 'previous', 'results'}:
    raise NoResultsError(response)
```

User will be able to write following code.
```python
client = PDCClient('dev')
try:
    for product in client.products:
        # ...
except NoResultsError as e:
    logger.error('Expected results, got: %s', e.response())
```

pdc_client/__init__.py
xiangge: Actually I haven't finish the patch yet. Will commit the new change here.

pdc_client/__init__.py
xiangge: good idea

xiangge:
@hluk  

> As I mentioned before, I don't like having to depend on a argument in constructor and to have the call function to return generator conditionally.

I think we should keep such argument, because currently people are using the codes now, we should keep the original behavior, such codes change the original logic, I think this is not correct.

```
client = PDCClient('dev')
try:
    for product in client.products:
        # ...
except NoResultsError as e:
    logger.error('Expected results, got: %s', e.response())
```

xiangge:
Patch committed. What's your thinking about it? @hluk

pdc_client/__init__.py
```diff
@@ -168,8 +165,12 @@ def decode(req):
 
         content_type = "application/json"
         encode = json.dumps
-        self.client = _BeanBagWrapper(beanbag.BeanBag(url, session=self.session, fmt=(content_type, encode, decode)),
-                                      page_size)
+
+        # In get_paged function it returns an interator, so user can get each pages quickly in plugins.
```
simozhan: typo: iterator

simozhan (approved):
LGTM

hluk:
@xiangge
> I think we should keep such argument, because currently people are using the codes now, we should keep the original behavior, such codes change the original logic, I think this is not correct.

I was taking about `omit_paging` argument which is not available in previous version so nobody uses it - I don't see any behavior being broken.

xiangge:
@hluk maybe I didn't say it clearly, will ping you directly :)

xiangge:
Code changed, please review that, I just let all the situations return a iterator and do the check in worker(), is that ok?

pdc_client/__init__.py
```diff
@@ -254,21 +251,48 @@ class _BeanBagWrapper(object):
...
+            try:
+                while True:
+                    response = self.client(*args, **kwargs)
+                    if isinstance(response, list):
+                        yield response
+                    else:
+                        yield response['results']
+                    if isinstance(response, dict) and response['next']:
+                        if response.keys() != ['count', 'previous', 'results', 'next']:
+                            raise Exception("No expected result %s" % response)
```
hluk: This should be specialized exception (`NoResultsError`?) which would allow user to process the unexpected response (so don't stringify it).

pdc_client/__init__.py
```diff
@@ -254,21 +251,48 @@ class _BeanBagWrapper(object):
...
+                        kwargs['page'] += 1
+                    else:
+                        break
+            except Exception as e:
+                raise e
```
hluk: Remove.

pdc_client/__init__.py
```diff
@@ -254,21 +251,48 @@ class _BeanBagWrapper(object):
...
+        def worker():
+            kwargs['page'] = 1
+            try:
+                while True:
+                    response = self.client(*args, **kwargs)
+                    if isinstance(response, list):
+                        yield response
+                    else:
+                        yield response['results']
+                    if isinstance(response, dict) and response['next']:
```
hluk: This can throw `KeyError`, the condition below should be checked first.

pdc_client/__init__.py
```diff
@@ -254,21 +251,48 @@ class _BeanBagWrapper(object):
...
     def __init__(self, client, page_size):
         self.client = client
         self.page_size = page_size
 
     def __call__(self, *args, **kwargs):
         if 'page_size' not in kwargs:
             kwargs['page_size'] = self.page_size
-        return self.client(*args, **kwargs)
+        result = self.client(*args, **kwargs)
+        return result
```
hluk: Just revert to the previous code.

hluk (changes requested):

pdc_client/__init__.py
xiangge: yes, rollback

pdc_client/__init__.py
xiangge: reomved

pdc_client/__init__.py
xiangge: Change the logic, and add a new NoResultsError here as there is no builtin type.

pdc_client/__init__.py
```diff
@@ -90,6 +90,9 @@ def read_config_file(server_alias):
     return result
 
 
+class NoResultsError(Exception):
+    pass
```
hluk: User should see details about what went wrong in the response (`{"detail":...}`, `{"bad_field":...}` or similar) so it should be accessible from the exception (e.g. `e.response().get('detail')`).

Also add docstrings here and to the `results()` method.

pdc_client/__init__.py
```diff
@@ -269,6 +270,31 @@ def __getattr__(self, name):
     def __getitem__(self, *args, **kwargs):
         return _BeanBagWrapper(self.client.__getitem__(*args, **kwargs), self.page_size)
 
+    def results(self, *args, **kwargs):
+        """
+           Return an iterator with all pages of data when multi pages.
+           Return a list when there is a single page.
```
hluk: This is misleading since the return type is always an iterator (not a list).

I think this must be documented elsewhere (e.g. usage examples in docstring for `PDCClient`), user shouldn't know that there is `_BeanBagWrapper` private class ([API docs for private stuff is not generated](https://product-definition-center.github.io/pdc-client/api.html)).

hluk (changes requested):

pdc_client/__init__.py
xiangge: oh “so don't stringify it”， I think you don't agree with add a string here.
This makes sense to me

pdc_client/__init__.py
xiangge: Yep, will change that.

pdc_client/__init__.py
xiangge: Have changed the docstring of results function. Add the example in PDCClient docstring.
And also add string in Exception, not sure if this is what your mean, if not, would you please give me an exactly example?

hluk (approved):
Thanks for your patience! :)

lubomir (approved):

xiangge:
@hluk it's ok, thanks for the advises of patch :).

PR #121, opened 2017-09-07
Fix installing plugins and running tests

hluk:
Few things happening here:
1. Merge branch 'release'
2. Revert removing comments from downstream
3. Improve installing plugins
    
    Install bytecode for plugins only for correct version of Python.
    
    Replaces default plugin path in code with real installation path to
    plugins.
    
    JIRA: PDC-2193
4. Fix running tests when building rpm


hluk:
I had to allow changing plugin path for pdc so the tests can be run when building RPM.

With this patch you can specify plugin directories.

```bash
PDC_CLIENT_PLUGIN_PATH="/path/to/plugins1:/path/to/plugins2" pdc release update --help
```

I remember this feature was mentioned by @ahills; not sure if there is JIRA for this.

pdc_client/runner.py
```diff
@@ -36,11 +36,16 @@ def autocomplete(*args):
...
+# install plugin path (usually '/usr/share/pdc-client/plugins').
+DEFAULT_PLUGIN_DIR = os.path.join(os.path.dirname(__file__), 'plugins')
+
 # A list of paths to directories where plugins should be loaded from.
 # The purpose of the plugins is to extend the default behaviour.
-LOCAL_DIR = os.path.join(os.path.dirname(__file__), 'plugins')
-INSTALLED_DIR = os.path.join('/usr/share/pdc-client', 'plugins')
-PLUGIN_DIRS = [LOCAL_DIR] if os.path.exists(LOCAL_DIR) else [INSTALLED_DIR]
+PLUGIN_DIRS = os.getenv('PDC_CLIENT_PLUGIN_PATH', '').split(':')
+PLUGIN_DIRS = [path for path in PLUGIN_DIRS if path]
```
xiangge: Seems these two lines are duplicate?

hluk: Just reusing the variable. Perhaps I should put it under one assignment and multiple lines.

xiangge: Maybe I didn't say my point clearly, I just meat there is no need to add the second line.

But I am wrong, the second line is used to remove the null chart like ' '. 
Cool! Then no question about this.

xiangge: And I think both one assignment with two lines or two assignments are ok.

hluk: Already changed it :). It makes sense to have only single assignment for each "constant".

simozhan (approved):
LGTM

xiangge (approved):
Looks good to me

PR #124, opened 2017-09-19
Fix printing errors and exit code for pdc_client

hluk:
Use a logger to print errors and debug messages.

Set non-zero exit code on error.

JIRA: PDC-2132

### DIFF
--- a/bin/pdc_client
+++ b/bin/pdc_client
@@ -12,6 +12,7 @@ except ImportError:
     import simplejson as json
 
 import argparse
+import logging
 import sys
 
 import requests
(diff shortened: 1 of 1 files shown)


simozhan (approved):
LGTM. Just one thing i'd like to know, what's the reason to replace 'with' statement(includes __enter__/__exit__) with finally? For simplicity?

hluk:
@simozhan
> what's the reason to replace 'with' statement(includes enter/exit) with finally?

I think using context manager (i.e. enter/exit) doesn't make sense here. Normally it's used to ensure that resources (files, sockets, etc.) are freed after an instance is not needed. But here it was used to store exception and print trace if needed which can be handled directly in try/except/finally.

xiangge:
@simozhan and @hluk maybe the original "with" block will automatically call the __exit__?

bin/pdc_client
```diff
@@ -173,12 +161,17 @@ if __name__ == "__main__":
...
         if options.comment:
             client.set_comment(options.comment)
+
+        request = RequestMethod(client, options.resource, logger)
+        result = request.dispatch(options.request, data)
+        print(json.dumps(result, indent=4, sort_keys=True))
     except BeanBagException as e:
-        print("%d %s" % (e.response.status_code, e.response.content))
+        logger.error("%d %s" % (e.response.status_code, e.response.content))
+        sys.exit(1)
```
xiangge, line 171: If we don't have a sys.exit(1) here, what will happend?

simozhan:
@hluk Yeah. The change makes sense. But maybe more commit message can make it clear, right? :)

hluk:
@simozhan 
>  But maybe more commit message can make it clear, right? :)

Amending to:

    Replaces `with` statement with try/except/finally and remove context
    manager. Context managers are normally used to free resources after an
    object is not needed, here it was abused to store exception and print
    stack.

hluk:
@xiangge 
> maybe the original "with" block will automatically call the exit?

Yes, `__exit__()` is called automatically at the end of `with` block (see [With Statement Context Managers](https://docs.python.org/2/reference/datamodel.html#context-managers)).

simozhan:
@hluk Yeah. That is really clear. Cool. Thanks!

bin/pdc_client
xiangge: Just realize if not add this line, will return 0 in $?.

xiangge (approved):
LGTM

xiangge:
@hluk yes, sorry, looked the wrong codes in my local env which __exit__ is sitll there :).

PR #127, opened 2017-09-20
Remove duplicate code

hluk:
Instead of patching test methods using meta class, which has
incompatible declarations between Python version 2 and 3, patch test
method when it's requested (using `__getattribute__()`).

---

To make reviewing code easier for you, the only thing I changed is removed meta class stuff and add following.

```python
class CLITestCase(unittest.TestCase):
    def __getattribute__(self, name):
        """
        Automatically passes mocked API object as argument to all test methods.
        """
        attr_value = super(CLITestCase, self).__getattribute__(name)

        if callable(attr_value) and name.startswith('test_'):
            def wrapper():
                with mock.patch('pdc_client.PDCClientWithPage') as cls:
                    api = MockAPI()
                    cls.return_value = api
                    return attr_value(api)
            return wrapper

        return attr_value
```


xiangge:
@hluk cool patch, but here is one warning when I do the test with "python3 setup.py test".

> test_info_json (tests.build_image.tests.BuildImageTestCase) ... /home/chcao/cc-dev/review/lukas/pdc-client/pdc_client/runner.py:113: ResourceWarning: unclosed file <_io.TextIOWrapper name='/home/chcao/cc-dev/review/lukas/pdc-client/pdc_client/plugins/repo.py' mode='r' encoding='utf-8'>
  file, pathname, description = imp.find_module(name[:-3], [dir])
/home/chcao/cc-dev/review/lukas/pdc-client/pdc_client/runner.py:113: ResourceWarning: unclosed file <_io.TextIOWrapper name='/home/chcao/cc-dev/review/lukas/pdc-client/pdc_client/plugins/compose.py' mode='r' encoding='utf-8'>

Just check with that, seems we should close the "file" obj of line 113 in runner.py after it is used.

hluk:
> here is one warning when I do the test with "python3 setup.py test".

@xiangge Umm, doesn't look like it's issue in my changes. I can see the error even on master. Can you verify?

xiangge:
@hluk yes, you are right, will file a bug for this.
This patch looks good to me.

xiangge (approved):

simozhan (approved):
LGTM

PR #128, opened 2017-09-22
Change the docstrings and fix issues

xiangge:
Improve the docstrings of pdc-client
Fix issues on page_size of constructor

JIRA: PDC-2120

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -98,18 +98,24 @@ class PDCClient(object):
     connections. The authentication token is automatically retrieved (if
     needed).
     """
+
     def __init__(self, server, token=None, develop=None, ssl_verify=None, page_size=None):
-        """Create new client instance.
+        """
+        Create new pdc client instance.
 
         Once the class is instantiated, use it as you would use a regular
-        BeanBag object. Please see its documentation to see how to use this
+        BeanBag object. Please look at its documentation for how to use this
         class to perform requests.
 
-        :param server:     server API url or server name from configuration
-        :paramtype server: string
+        :param server:     Server API url or server name from configuration
+        :param token:      An authentication token string of visiting pdc server
+        :param develop:    This is use for dev mode
         :param ssl_verify: True for validating SSL certificates with system CA
                            store; False for no validation; path to CA file or
                            directory to use for validation otherwise
+        :param page_size:  This is a number of data which is returned per page.
+                           A -1 means that pdc server will return all the data in
+                           one request.
         """
         self.page_size = page_size
         if not server:
@@ -162,7 +168,8 @@ class PDCClient(object):
 
         content_type = "application/json"
         encode = json.dumps
-        self.client = beanbag.BeanBag(url, session=self.session, fmt=(content_type, encode, decode))
+        self.client = _BeanBagWrapper(beanbag.BeanBag(url, session=self.session, fmt=(content_type, encode, decode)),
+                                      page_size)
         if not develop:
             # For develop environment, we don't need to require a token
             if not token:
@@ -238,6 +245,29 @@ class PDCClient(object):
         :paramtype comment: string
         """
         self.session.headers["PDC-Change-Comment"] = comment
+
+
+class _BeanBagWrapper(object):
+    """
+       Wrapper of BeanBag's attributes and items.
+
+       This class wraps attributes and items of Beanbag to let page_size of
+       PDCClient's constructor work.
+    """
+    def __init__(self, client, page_size):
+        self.client = client
+        self.page_size = page_size
+
+    def __call__(self, *args, **kwargs):
+        if 'page_size' not in kwargs:
+            kwargs['page_size'] = self.page_size
+        return self.client(*args, **kwargs)
+
+    def __getattr__(self, name):
+        return _BeanBagWrapper(self.client.__getattr__(name), self.page_size)
+
+    def __getitem__(self, *args, **kwargs):
+        return _BeanBagWrapper(self.client.__getitem__(*args, **kwargs), self.page_size)
 
 
 class PDCClientWithPage(PDCClient):


pdc_client/__init__.py
```diff
@@ -91,27 +91,71 @@ def read_config_file(server_alias):
...
+               :Example:
+
+               >>> x = BeanBag("http://host/api")
+               >>> r = x()                                 # GET request
+               >>> r = x(p1='foo', p2=3)                   # GET request with parameters passed via query string
+               >>> r = x( {'a': 1, 'b': 2} )               # POST request
+               >>> r = x( "RANDOMIZE", {'a': 1, 'b': 2} )  # Custom HTTP verb with request body
+               >>> r = x( "OPTIONS", None )                # Custom HTTP verb with empty request body
+            """
+            if PDCClient.page_size is not None:
```
hluk: `PDCClient.page_size` is static, you need `page_size` for given instance of `PDCClient`.

pdc_client/__init__.py
```diff
@@ -91,27 +91,71 @@ def read_config_file(server_alias):
...
+
+    class _WrapperBeanBag(beanbag.BeanBag):
+        """
+        BeanBag wrapper for getting the constructor's page_size in used.
+
+        Re-implement the call function as in beanbag you can't call
+        super() function of parent class.
+        """
+
+        def call(self, path, *args, **kwargs):
```
hluk: Don't copy-paste the method, use `super().call(...)` and docstring should just explain why is it overridden.

pdc_client/__init__.py
```diff
@@ -91,27 +91,71 @@ def read_config_file(server_alias):
 
 
 class PDCClient(object):
-    """BeanBag wrapper specialized for PDC access.
```
hluk: I sometimes do this too, but try to change only relevant lines for the change. It's easier to review and "git blame" gives some sane results.

pdc_client/__init__.py
```diff
@@ -91,27 +91,71 @@ def read_config_file(server_alias):
...
+    BeanBag wrapper specialized for PDC access.
 
     This class wraps general BeanBag.v1 objects, but provides easy-to-use
     interface that can use configuration files for specifying server
     connections. The authentication token is automatically retrieved (if
     needed).
     """
+
+    # Set a page_size which is used in _WrapperBeanBag
+    page_size = None
```
hluk: I'm assuming this should be `DEFAULT_PAGE_SIZE`? This is probably not needed.

hluk (changes requested):

pdc_client/__init__.py
xiangge: This can't work, you can have a try in local env.

pdc_client/__init__.py
xiangge: Yes, I just want to use this static parameter. As I don't want to copy the __init__ function again.

pdc_client/__init__.py
xiangge: I just want to use this static parameter in wrapper class, and the DEFAULT_PAGE_SIZE name is more better, will change that.

pdc_client/__init__.py
xiangge: Make sense to me, will change this back, just think that write as this line is very ugly.
Don't why write like this although I saw beanbag also write the first line of docstrings like this.

pdc_client/__init__.py
xiangge: And when you try use a super().func, will raise an error like the parent class is something like _WrapperBeanBagBase but not BeanBag, so super not works here. I saw some changes in deep of the BeanBag's code.

pdc_client/__init__.py
hluk: Oops, sorry, didn't notice that it's the static variable. `DEFAULT_PAGE_SIZE` is for global constant.

xiangge: np, I think so, didn't use the static variable in class for some time.
Change back, but default_page_size is still good than page_size.

pdc_client/__init__.py
hluk: OK, I think I understand. You use the derived class only as `self.client` but there is no documentation and certainty about type returned in `__getitem__()` and `__getattr__()`.

If that's the case then overriding methods of `BeanBag` is not way to go.

I think you have to wrap return values of `__getitem__()` and `__getattr__()` so you can override (as I mentioned [here](https://github.com/product-definition-center/pdc-client/pull/119#issuecomment-328505577)).

pdc_client/__init__.py
hluk: Don't use static member in the instance, this will lead to bugs if you use multiple `PDCClient` instances.

pdc_client/__init__.py
```diff
@@ -98,20 +98,29 @@ class PDCClient(object):
     connections. The authentication token is automatically retrieved (if
     needed).
     """
+
+    # Set a default_page_size which is used in _WrapperBeanBag
```
hluk: Remove.

pdc_client/__init__.py
```diff
@@ -240,6 +250,29 @@ def set_comment(self, comment):
         self.session.headers["PDC-Change-Comment"] = comment
 
 
+class BeanBagAttrWrapper(object):
```
hluk: `_BeanBagWrapper` was perhaps better.

pdc_client/__init__.py
```diff
@@ -98,20 +98,29 @@ class PDCClient(object):
...
+        :param develop:    This is use for dev mode
         :param ssl_verify: True for validating SSL certificates with system CA
                            store; False for no validation; path to CA file or
                            directory to use for validation otherwise
+        :param page_size:  This is a number of data which is returned per page.
+                           A -1 means that pdc server will return all the data in
+                           one request.
         """
         self.page_size = page_size
+        PDCClient.default_page_size = page_size
```
hluk: Remove.

hluk (changes requested):

hluk (approved):

simozhan (approved):
LGTM

PR #131, opened 2017-09-28
Add the close function when load plugins

xiangge:
Add the close function when load plugins
JIRA: PDC-2235

### DIFF
--- a/pdc_client/runner.py
+++ b/pdc_client/runner.py
@@ -110,14 +110,21 @@ class Runner(object):
             for name in os.listdir(dir):
                 if not name.endswith('.py') or name not in plugins_set:
                     continue
-                file, pathname, description = imp.find_module(name[:-3], [dir])
-                plugin = imp.load_module(name[:-3], file, pathname, description)
-                self.logger.debug('Loaded plugin {0}'.format(name[:-3]))
-                self.raw_plugins.append(plugin)
-                if hasattr(plugin, 'PLUGIN_CLASSES'):
-                    for p in plugin.PLUGIN_CLASSES:
-                        self.logger.debug('Instantiating {0}'.format(p.__name__))
-                        self.plugins.append(p(self))
+                try:
+                    module_name = name[:-3]
+                    file, pathname, description = imp.find_module(module_name, [dir])
+                    plugin = imp.load_module(module_name, file, pathname, description)
+                    self.logger.debug('Loaded plugin {0}'.format(module_name))
+                    self.raw_plugins.append(plugin)
+                    if hasattr(plugin, 'PLUGIN_CLASSES'):
+                        for p in plugin.PLUGIN_CLASSES:
+                            self.logger.debug('Instantiating {0}'.format(p.__name__))
+                            self.plugins.append(p(self))
+                except Exception as e:
+                    self.logger.error('Failed to load plugin "{0}": {1}'.format(module_name, e))
+                finally:
+                    if file:
+                        file.close()
 
     def run_hook(self, hook, *args, **kwargs):
         """


pdc_client/runner.py
```diff
@@ -110,14 +110,19 @@ def load_plugins(self):
...
+                    file, pathname, description = imp.find_module(name[:-3], [dir])
+                    plugin = imp.load_module(name[:-3], file, pathname, description)
+                    self.logger.debug('Loaded plugin {0}'.format(name[:-3]))
+                    self.raw_plugins.append(plugin)
+                    if hasattr(plugin, 'PLUGIN_CLASSES'):
+                        for p in plugin.PLUGIN_CLASSES:
+                            self.logger.debug('Instantiating {0}'.format(p.__name__))
+                            self.plugins.append(p(self))
+                except:
+                    print('Failed to init plugin %s.' % name[:-3])
```
simozhan: 1. Maybe 'Failed to load plugin'
2. Is it ok to make consistent style with other log, like 'Loaded plugin {0}'.format...

simozhan (changes requested):

pdc_client/runner.py
xiangge: Changed, please review

simozhan (approved):

pdc_client/runner.py
```diff
@@ -110,14 +110,19 @@ def load_plugins(self):
...
+                    file, pathname, description = imp.find_module(name[:-3], [dir])
+                    plugin = imp.load_module(name[:-3], file, pathname, description)
+                    self.logger.debug('Loaded plugin {0}'.format(name[:-3]))
+                    self.raw_plugins.append(plugin)
+                    if hasattr(plugin, 'PLUGIN_CLASSES'):
+                        for p in plugin.PLUGIN_CLASSES:
+                            self.logger.debug('Instantiating {0}'.format(p.__name__))
+                            self.plugins.append(p(self))
+                except:
+                    print('Failed to load plugin {0}'.format(name[:-3]))
```
hluk: Omit `print()` for status and error messages. And try to show more information.
```python
except Exception as e:
    self.logger.error('Failed to load plugin "{0}": {1}'.format(name[:-3], e))
```

pdc_client/runner.py
```diff
@@ -110,14 +110,19 @@ def load_plugins(self):
...
+                finally:
+                    file.close()
```
hluk: Use `if file:` here.

pdc_client/runner.py
```diff
@@ -110,14 +110,19 @@ def load_plugins(self):
...
-                file, pathname, description = imp.find_module(name[:-3], [dir])
-                plugin = imp.load_module(name[:-3], file, pathname, description)
-                self.logger.debug('Loaded plugin {0}'.format(name[:-3]))
-                self.raw_plugins.append(plugin)
-                if hasattr(plugin, 'PLUGIN_CLASSES'):
-                    for p in plugin.PLUGIN_CLASSES:
-                        self.logger.debug('Instantiating {0}'.format(p.__name__))
-                        self.plugins.append(p(self))
+                try:
+                    file, pathname, description = imp.find_module(name[:-3], [dir])
```
hluk:
```python
module_name = name[:-3]
```

hluk (changes requested):

pdc_client/runner.py
xiangge: Changed.

pdc_client/runner.py
xiangge: I think use print here more makes sense, we have no need to show a "Error:pdc: ......" when ct meets error in command lines, but the error info should be included.
Changed.

pdc_client/runner.py
xiangge: changed, but do you think it's possible to get a None value of file here?

pdc_client/runner.py
hluk: With `print()` you're printing on stdout by default, I wouldn't expect any errors there. Let logger handle the details. Also it makes messages to follow some format and I can grep errors easily.

xiangge: # Also it makes messages to follow some format and I can grep errors easily.
Make sense, change

xiangge: Changed, please help to review, will commit another patch to change other print

hluk (approved):

PR #132, opened 2017-10-16
Add comments for rpm requirements

xiangge:


requirements/production.txt
```diff
@@ -1,3 +1,4 @@
 requests
+# requires: krb5-devel kobo-rpmlib
```
lubomir: Kobo should be installable directly from PyPI. Maybe just list it as a requirement?

xiangge: Test it again, no kobo-rpmlib dependency here, remove this part.

requirements/production.txt
```diff
@@ -1,3 +1,4 @@
 requests
+# requires: krb5-devel
```
hluk, line 2: Is this required and installed by python-pdc package? The dependency is not explicitly mentioned in spec file (only in Docker).

Also [python-requests-kerberos.spec](https://src.fedoraproject.org/rpms/python-requests-kerberos/blob/f26/f/python-requests-kerberos.spec) doesn't explicitly require it for `python2-requests-kerberos` and `python3-requests-kerberos` packages.

xiangge: In pdc-server'd codes, there is no code used requests_kerberos, so no such comment and requirements here.
And if you use yum install , it can be success, but use a pip/pip3 install, it needs this, seems need some related codes.

hluk (approved):

simozhan (approved):

PR #135, opened 2017-10-23
Fix wrapping BeanBag methods and operators

### LINKED ISSUE #134: Current client breaks BeanBag syntax sugar for updates
Using the syntactic sugar for PATCH request gives an exception:

```python
> c = pdc_client.PDCClient('local')
> c['unreleasedvariants']['389-ds-1.2-20171009091843/'] += {'active': False}
TypeError: unsupported operand type(s) for +=: '_BeanBagWrapper' and 'dict'
```

Implementing the `__iadd__` method on `_BeanBagWrapper` similarly to existing methods makes the PATCH request work, but once that request finishes, another exception is reported:

```
TypeError: '_BeanBagWrapper' object does not support item assignment
```

For the PATCH use case this is harmless, but it indicates somehow a PUT request was attempted.  Indeed running `c['unreleasedvariants']['389-ds-1.2-20171009091843/'] = {'active': False}` (data is invalid, but it doesn't matter) reports the same problem and does not perform the request.

Adding a `__setitem__` fixes the PUT request.

However if both methods exist, issuing a PATCH with the first example fires a PATCH request followed immediately by PUT request that should not be there.

hluk:
Wrap following BeanBag methods and operators:
- 'del' (DELETE)
- '=' (PUT)
- '+=' (PATCH)
- '=='

Fixes #134

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -113,7 +113,24 @@ class NoResultsError(Exception):
         return msg
 
 
-class PDCClient(object):
+class _SetAttributeWrapper(object):
+    """
+    Passes all __setattr__() calls to a client after _initialized is True.
+
+    Attribute client must be defined in derived class.
+
+    Set _initialized to True after all attributes for derived object are set.
+    """
+    _initialized = False
+    client = None
+
+    def __setattr__(self, name, value):
+        if self._initialized:
+            return self.client.__setattr__(name, value)
+        return super(_SetAttributeWrapper, self).__setattr__(name, value)
+
+
+class PDCClient(_SetAttributeWrapper):
     """BeanBag wrapper specialized for PDC access.
 
     This class wraps general BeanBag.v1 objects, but provides easy-to-use
@@ -216,6 +233,8 @@ class PDCClient(object):
                 token = self.obtain_token()
             self.session.headers["Authorization"] = "Token %s" % token
 
+        self._initialized = True
+
     def obtain_token(self):
         """
         Try to obtain token from all end-points that were ever used to serve the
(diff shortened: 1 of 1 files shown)


pdc_client/__init__.py
```diff
@@ -305,9 +329,24 @@ def __call__(self, *args, **kwargs):
...
+
+    def __delitem__(self, name):
+        return self.client.__delitem__(name)
+
+    def __iadd__(self, value):
+        return self.client.__iadd__(value)
+
+    def __eq__(self, other):
+        return self.client == other.client
+
```
xiangge, line 349: Do we need to add this eq function? Beanbag codes don't init this function.
There are also a str and path function here, should think about if we need re-write them.

BTW: This wrapper class makes us wrapper all the functions.

lubomir (approved):
This looks good to me. The original issue can not be replicated with this patch and I can't find any other case where this would not work.

pdc_client/__init__.py
hluk: Nice catch. I've added `__str__()`.

Method for equality check is needed - there is [an example for it in BeanBag API docs](http://beanbag.readthedocs.io/en/latest/v1.html).

xiangge: ok, cool

pdc_client/__init__.py
```diff
@@ -216,6 +233,8 @@ def decode(req):
                 token = self.obtain_token()
             self.session.headers["Authorization"] = "Token %s" % token
 
+        self._initialized = True
+
```
xiangge, line 237: I think we don't need to set the base class to _SetAttributeWrapper in PDCClient.

As all the attrs or items we get from PDCClient are now from Beanbag, so after that we just need to wrap the set function on _BeanBagWrapper.

What's your thinking?

hluk: Method `__setattr__()` needs to be overridden in `PDCClient` to handle code like `client.resource = {...}` correctly.

xiangge: Make sense, if people set the resource url as server, like v1/releases/, the setattr should start from client itself.

xiangge (approved):

simozhan (approved):

PR #136, opened 2017-10-24
Let attr work perfect

xiangge:
client.base-products can't work, let client.base_products instead.

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -288,6 +288,20 @@ class PDCClient(_SetAttributeWrapper):
         return self.client(*args, **kwargs)
 
     def __getattr__(self, name):
+        """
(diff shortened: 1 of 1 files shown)


hluk (approved):

simozhan (approved):
LGTM

lubomir:
Admittedly dashes are more common than underscores, but there can be items with user supplied identifier that contain underscores. I expect global/release component names could quite easily contain underscores. So this is not perfect.

xiangge:
@lubomir  
Yes, this patch can just solve some issues but not all, currently pdc's resources include "-" and this patch just covers first layer resource name. We can't replace all because the columns include "_".

I think beanbag itself should do some works.

PR #137, opened 2017-10-25
Add PDCClient tests

hluk:


hluk:
Keeping compatibility with Python 2.6 sucks - still have to fix some tests.

xiangge:
Yes, I think we can consider to remove the 2.6 test in the future and add update the 3.3 to 3.5.

1. And also I think in the case we should include the attr "_" to add a "/" in the end of the uri.
2. And we should also add the getitem way in each of the test function, like client["cpes"].

hluk:
While adding tests for rest of the API I found few issues. Fixes are included.

> 1. And also I think in the case we should include the attr "_" to add a "/" in the end of the uri.

Added.

> 2. And we should also add the getitem way in each of the test function, like client["cpes"].

This is already tested on few places. I want to have simple and small test methods for testing just specific part of API.

hugovk:
Here's the pip installs for pdc-client from PyPI for the last month -- a single 2.6 and no 3.3 (both are EOL):
```
$ pypinfo --percent --pip pdc-client pyversion
python_version percent download_count
-------------- ------- --------------
2.7              63.9%            991
3.6              20.6%            319
3.5              12.5%            194
3.4               2.9%             45
2.6               0.1%              1

```

By the way, you could add both 3.5 and 3.6 to Travis etc. More reasons for dropping old ones:

### 2.6

* Unsupported since [2013-10-29](https://en.wikipedia.org/wiki/CPython#Version_history)
* https://snarky.ca/stop-using-python-2-6/
* http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html
* http://www.python3statement.org
* Current pip 9 deprecates Python 2.6 support, pip 10 won't support it (https://github.com/pypa/pip/issues/3955)
* Not much PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

### 3.2

* (Already dropped! :)

### 3.3

* Unsupported since [2017-09-29](https://en.wikipedia.org/wiki/CPython#Version_history)
* pip 10 will deprecate Python 3.3 support, pip 11 won't support it (https://github.com/pypa/pip/issues/3796)
* Very little PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

tests/api/tests.py
```diff
@@ -0,0 +1,381 @@
...
+            "allowed_push_targets": []
+        }
+    }
+    cpes = {
+        1: {
+            "cpe": "cpe:/o:redhat:enterprise_linux:7::workstation",
+            "description": "RHEL 7 Workstation",
+            "id": 1
+        }
+    }
```
xiangge: Is the cpe returned data in our product something like below one?
 [
        {
            "id": 1,
            "cpe": "cpe:test",
            "description": "1234"
        }
    ]

tests/api/tests.py
```diff
@@ -0,0 +1,381 @@
...
+        self.assertEqual(self.client.products, self.client.products)
+        self.assertEqual(self.client['products'], self.client['products'])
+        self.assertEqual(self.client.products, self.client['products'])
+        self.assertEqual(self.client.products.fedora, self.client['products']['fedora'])
+        self.assertEqual(self.client._.products.fedora._, self.client['products/']['fedora/'])
+
+    def test_set_comment(self):
+        self.client.set_comment('TEST')
+        self.client.cpes[1] += {'description': 'TEST'}
+        self.assertEqual(_MockPDCServerRequestHandler.last_comment, 'TEST')
```
xiangge, line 412: > And we should also add the getitem way in each of the test function, like client["cpes"].

> This is already tested on few places. I want to have simple and small test methods for testing just specific part of API.
 
I think we can add codes here to test patch as item, not mean add more test functions.
```
 self.client['cpes'][1] += {'description': 'TEST1'}
 self.assertEqual(_MockPDCServerRequestHandler.last_comment, 'TEST1')
```
Maybe you mean self.client.cpes[1] is the case that gets resource by item "[1]"?

xiangge:
I just found I keep the comment there for several days, but not submit it,sorry about that...
Why github always let me submit again after I give comment ...

tests/api/tests.py
```diff
@@ -0,0 +1,381 @@
...
+        # Python 2.6 compatibility
+        if not hasattr(PDCClientTestCase, 'url'):
+            PDCClientTestCase.setUpClass()
+
+        self.client = PDCClient(
+            server=self.url,
+            ssl_verify=False,
+        )
+
+    def test_request_response(self):
```
simozhan: I don't quite understand this case. Is this testing the Mock server setup correctly? I think other cases already cover this.

simozhan (approved):
Nice patch. Just one minor thing related to the test case

tests/api/tests.py
hluk: The mocked server returns paged response when making GET HTTP request to `rest_api/v1/cpe/`.

tests/api/tests.py
hluk: OK, I'll add that. Also patching using `+=` for attribute would be nice.

tests/api/tests.py
hluk: OK, I'll remove it. It was the first test just to verify the server is running.

hluk:
I simplified the server code a lot (no regexps and should be easier to modify).

I also added few more tests for PUT and PATCH.

hluk:
[Test coverage for the last change](https://coveralls.io/builds/14174424/source?filename=pdc_client%2F__init__.py)

pdc_client/__init__.py
```diff
@@ -361,17 +369,14 @@ def worker():
                 response = self.client(*args, **kwargs)
                 if isinstance(response, list):
                     yield response
-                else:
+                elif _is_page(response):
```
xiangge: If you change like this, it seems that when the response is a list, the "while true" seems not break?

tests/api/tests.py
xiangge: So the data is like the data in db, got that, thx.

simozhan:
Reivewed again. LGTM

tests/api/tests.py
xiangge: cool

pdc_client/__init__.py
hluk: Nice catch. Fixed and added test for this.

xiangge: cool

xiangge (approved):

PR #138, opened 2017-11-17
Modify and merge create-release.sh to pdc-client

xiangge:


create-release.sh
```diff
@@ -0,0 +1,49 @@
...
+set -o pipefail
+
+fail() {
+    echo "Error: $*" 1>&2
+    exit 1
+}
+
+current_branch=$(git rev-parse --abbrev-ref HEAD)
+
+if [[ "$current_branch" != "release" ]]; then
```
simozhan: I am guessing if you have different remotes, this will fail.
In my local env, I will use "git checkout release origin/release" to get the origin's release branch.
Maybe simply fail is better, let the user to checkout the correct release branch.

create-release.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo 'Checkout "release".'
+    git checkout release
+fi
+
+if [[ "$1" == "skip-merge" ]]; then
```
simozhan: I think this "git merge master" is ok for the first time. But if we have some fix in release and master branch at the same time with some conflicts, there will be some problem.

create-release.sh
```diff
@@ -0,0 +1,49 @@
...
+if [[ ! "$release" =~ ^[1-9][0-9]*$ ]]; then
+    fail 'Unexpected release format.'
+fi
+
+sed -i 's/\(^Version:\s*\).*/\1'"$version"'/' pdc-client.spec
+sed -i 's/\(^Release:\s*\).*/\1'"$release"'%{?dist}/' pdc-client.spec
+
+tito tag --keep-version
+
+echo Do a test build here
```
simozhan: I am thinking this script is only for adding a new commit with tito, not including tito build

simozhan (commented):
I think maybe we can simply add the same file as the server's. My idea is we provide the minimum script for automatically adding a commit with tito, we don't need to provide tito build in this script.

create-release.sh
xiangge: As we will upload srpm for rhpkg, I am thinking to add this command to make a srpm at this time. 
Remove this also make sense to me, just want to make it auto execute.

create-release.sh
xiangge: Yes, correct, maybe just give an error info like original codes.
In my local env, I just want to fork a master in new folder, then all the things do automatically :).

create-release.sh
xiangge: Yes, this is why I add a skip-merge to let people choose if they do it manually or not.
I think in your server, you will do with "skip-merge" as there are many remotes there.

create-release.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+echo "Current git commit is: $(git describe)"
+
+read -p 'Enter new version (e.g. "1.8.0"): ' version
+if [[ ! "$version" =~ ^[1-9][0-9]*\.[0-9]+\.[0-9]+$ ]]; then
+    fail 'Unexpected version format.'
+fi
+
+read -p 'Enter new release number: ' release
```
lubomir: You should always use `-r` option with read: https://github.com/koalaman/shellcheck/wiki/SC2162
It's not going to cause problems here though.

lubomir (commented):
The script looks good. The release checklist should be updated to use this, right?

create-release.sh
xiangge: Yes, use this script can modify spec and also create new srpm for bew build.

The "read -p" is copied from the pdc-server repo's codes it can make the echo and read in one line.
Like you said here we verify the input, the version just can be "x.y.z", release just can be a number. Both "read -p" and "echo / read -r" should work fine.  But I have changed to read -r to keep good use.

create-release.sh
xiangge: I add  codes to let ct choose if they want to checkout release automatically.

create-release.sh
xiangge: Also add codes to let ct choose if they want to merge master to release automatically.

hluk (approved):

simozhan (approved):

PR #140, opened 2017-11-24
Release

xiangge:
If use the new setup.py test -q in spec file, will raise below error as there is no tests folder in tar.gz.

> ....
> romNames
>     suites = [self.loadTestsFromName(name, module) for name in names]
>   File "/usr/lib64/python2.7/unittest/loader.py", line 91, in loadTestsFromName
>     module = __import__('.'.join(parts_copy))
> ImportError: No module named tests


pdc-client.spec
```diff
@@ -142,11 +144,11 @@ find -name '*.py' | xargs sed -i '1s|^#!python|#!%{__python}|'
 export PDC_CLIENT_PLUGIN_PATH="%{buildroot}%{plugin_install_path}"
 test -d "$PDC_CLIENT_PLUGIN_PATH"
 
-%{__python2} setup.py test -q
+%{__python2} setup.py nosetests
```
simozhan: Can we totally remove this setup.py test -q ( or nosetests)?
I am a bit lost why we need this test in spec file.
Seems nosetests will not raise error when test module doesn't exist.

simozhan (commented):
Comment in line

pdc-client.spec
xiangge: Yes, I agree, from the tito build we can see there is no tests executed because there is no tests folder even use a nose.

> writing manifest file 'pdc_client.egg-info/SOURCES.txt'
> 
> ----------------------------------------------------------------------
> Ran 0 tests in 0.064s
> 
> OK

hluk: > Can we totally remove this setup.py test -q ( or nosetests)?

Please don't, I added this so it can reveal some problems before QE testing.

@xiangge What was the error with "setup.py test -q"?

xiangge: @hluk when I did the "tito build --rpm --offline" it raise this error.
And also when I did the "rhpkg build --scratch" also failed.
But when I change the nose back to the spec or remove it, there is no error comes.

And also when we do a build, the tests folder will not include in, so error happens.

hluk: I had a video chat with @xiangge and we found out that I can run the tests because I have PYTHONPATH variable set to a parent directory with the test (that's why it worked for me). Sorry about that.

We can remove the test invocation from spec file for this release but I would like to have that available for next releases since it can really safe a lot of time.

Thanks @xiangge for your time and sorry to keep you at work so late. :)

xiangge: @hluk thanks for your time for discussion :). and I will also do some investigation about test.

@simozhan and @hluk I have remove the test lines, please review.

pdc-client.spec
```diff
@@ -142,11 +142,8 @@ find -name '*.py' | xargs sed -i '1s|^#!python|#!%{__python}|'
 export PDC_CLIENT_PLUGIN_PATH="%{buildroot}%{plugin_install_path}"
 test -d "$PDC_CLIENT_PLUGIN_PATH"
 
-%{__python2} setup.py test -q
-
 %if 0%{?with_python3}
```
simozhan: I think we can remove the following several lines

simozhan (commented):
Comment in line

pdc-client.spec
xiangge: cool! that's correct, removed, thx @simozhan

simozhan (approved):

hluk (approved):

lubomir (commented):
I'm confused: the changelog talks about change in BuildRequires, but there is no such change in the commit.

xiangge:
@lubomir I rollback the nose BuildRequires in the first change, but test with that the nose tool is never  used, so remove them again in the second change. 
Will merge the two commit log to one, thanks for the comment.

PR #141, opened 2017-11-28
Rollback codes on parent class of PDCClient

xiangge:
Let PDCClient's parent class be object, but
not _SetAttributeWrapper, then we can set PDCClient's
attrs after or at the end of the __init__ function,
but not a BeanBag's attr.

JIRA: PDC-2429

### DIFF
--- a/pdc_client/__init__.py
+++ b/pdc_client/__init__.py
@@ -135,7 +135,7 @@ class _SetAttributeWrapper(object):
         return super(_SetAttributeWrapper, self).__setattr__(name, value)
 
 
-class PDCClient(_SetAttributeWrapper):
+class PDCClient(object):
     """BeanBag wrapper specialized for PDC access.
 
     This class wraps general BeanBag.v1 objects, but provides easy-to-use
@@ -237,8 +237,6 @@ class PDCClient(_SetAttributeWrapper):
             if not token:
                 token = self.obtain_token()
             self.session.headers["Authorization"] = "Token %s" % token
-
-        self._initialized = True
 
     def obtain_token(self):
         """
--- a/tests/api/tests.py
+++ b/tests/api/tests.py
@@ -387,9 +387,6 @@ class PDCClientTestCase(unittest.TestCase):
         with self.assertRaises(BeanBagException):
             self.client['bad_resource'] = {}
 
-        with self.assertRaises(BeanBagException):
-            self.client.bad_resource = {}
-
     def test_str(self):
         self.assertEqual(str(self.client.products.fedora), self.url + '/products/fedora')
         self.assertEqual(str(self.client), self.url + '/')


xiangge:
The issue occurred is because that we set a PDCClient's page after set the super call in PDCClientWithPage, the _SetAttributeWrapper makes the page to BeanBag's page, that's not correct.

And also we can move the "self.page = page" before the super call, but I think this is not a good way, maybe in ct's codes, they also already initialized some attrs for PDCClient, but not for BeanBag.

And also from the original codes, there is a re-write __getitem__ / __setitem__ / __getattr__ , but not a re-write __setattr__, I think they considered this situation, if ct wants to set a PDCClient's attr they use __setattr__ and if they want to set BeanBag's resource, they use __setitem__.

xiangge:
@hluk @lubomir would you please help to review it?

pdc_client/__init__.py
```diff
@@ -135,7 +135,7 @@ def __setattr__(self, name, value):
         return super(_SetAttributeWrapper, self).__setattr__(name, value)
 
 
-class PDCClient(_SetAttributeWrapper):
```
hluk, line 138: But removing this functionality breaks API.

From commit description I don't see what the issue is. What PDCClient attributes cannot be set? Can you write a test case for it in "tests/api/tests.py"?

hluk (changes requested):

lubomir (changes requested):
As far as I understand it, the problem is that `PDCClientWithPage()` fails when trying to set the `page` attribute (just creating the instance is enough to replicate).

Could that instead be solved by first setting the `page` attribute and only then calling `__init__` on parent?

pdc_client/__init__.py
lubomir: I agree that it would be nice to have a test for this.

xiangge:
@lubomir yes, correct, as I mentioned we can do this to solve the issue.
But if someone want to set a attribute just for PDCClient after they init the PDCClient object. The issue will occur then, like below:
```
#!/usr/bin/python
import sys
from pdc_client import PDCClient

server = "upstream"
client = PDCClient(server)

client.aaa = 1234

results = []
for r in client["releases"].results():
    results.append(r)

print(results)
```


The original code can support this, because the parent is object (class PDCClient(object)).
Add this _SetAttributeWrapper in PDCClient will break previous codes.

pdc_client/__init__.py
xiangge: Yes, like @lubomir mentioned, the reason why it raise "404 error" is there is a self.page set after the super call.
And I think remove this don't break the API, because in previous codes, we can support codes like "client.aaa".
@hluk and @lubomir  I can add the test on codes, good idea.

lubomir:
> client.aaa = 1234

That should submit a `PUT` request, so people should not be able to set attributes on the client object, right?

xiangge:
@lubomir in original codes, people can set like this. Just as an attribute of PDCClient, it is not a PUT request.

pdc_client/__init__.py
xiangge: @lubomir and @hluk I have added the test case.
change the port to 8278, as it raises "socket.error: Address already in use".

@hluk  I think we should still keep this _SetAttributeWrapper, because we should set the "self.client = client" and "self.page_size = page_size" in wrapper class then we could re-write the __setattr__,

tests/api/tests.py
```diff
@@ -33,7 +33,7 @@
 DEFAULT_SERVER = 'localhost'
 
 PORT_ENV_VAR_NAME = 'PDC_CLIENT_TEST_SERVER_PORT'
-DEFAULT_PORT = 8378
+DEFAULT_PORT = 8278
```
hluk: Something wrong with port 8378?

tests/api/tests.py
```diff
@@ -387,8 +387,11 @@ def test_bad_put(self):
         with self.assertRaises(BeanBagException):
             self.client['bad_resource'] = {}
 
-        with self.assertRaises(BeanBagException):
-            self.client.bad_resource = {}
+    def test_set_attr(self):
+        # this attr is just for PDCClient's object
+        # not wrapper to beanbag's attr
+        self.client.any_attr = "This is a test attr"
+        self.assertEqual(self.client.any_attr, "This is a test attr")
```
hluk: No need to test this.

xiangge: I saw you and Lubomir think to add a test?  Even I don't think we should add one case for this patch, which test case should we add?

tests/api/tests.py
xiangge: Interesting, I copied the tests.py with tests-123.py, then the tests executed two times.
Sorry, my bad, rollback, :)

hluk: The test process was probably not closed correctly. The thread with fake server can use some improvements.

hluk (approved):

tests/api/tests.py
xiangge: Yes, I think so.

lubomir (approved):
It does fix the reported problem, so +1 from me.

xiangge:
@lubomir and @hluk  Thank you guys :).

PR #143, opened 2017-12-01
Add the page_size=None when get auth

xiangge:
Add the page_size=None when get auth, then
the call function of wrapper can't set the
page_size again.

JIRA: PDC-2446


hluk (approved):
Nice solution!

lubomir (approved):
Clever!

xiangge:
@hluk and @lubomir  Thank you very much for the quick response.
And @simozhan thanks a lot for your discussion with me.

PR #147, opened 2018-01-22
Really fix deployment of documentation

lubomir:
The --track option only works when creating a new branch. What the script does is set the remote tracking branch. Using --set-upstream-to this should work repeatedly, even when already set.


hluk (approved):
OK. Perhaps documentation should be build (without deployment) even for PRs.

Any reason PDC is not using readthedocs.org?

lubomir:
Good idea about RTD. I tried setting it up: http://pdc-client.readthedocs.io/en/latest/
I'M not sure if the automatic update is working though.

hluk:
Actually, I forgot that the github account is not connected to the readthedocs one, so we don't have the admin access there by default.

lubomir:
I tried creating the webhook on GitHub side.

PR #148, opened 2018-01-24
Run tests on Python 3.5 and 3.6

lubomir:


xiangge:
Maybe we can consider to remove the 3.3 or 3.4?

hluk:
> Maybe we can consider to remove the 3.3 or 3.4?

Also Fedora 25 has Python 3.6 so perhaps there is no need for 3.5 either?

**Edit:** Sorry meant to say "Fedora 26 has Python 3.6".

hluk:
"JIRA: PDC-2075" is missing in the commit message.

lubomir:
I updated the commit message and remove 3.3. EPEL 7 has Python 3.4, so I would prefer to keep that. We currently do not build a python 3 version for EPEL, but we should.

hluk (approved):

xiangge:
Cool

PR #151, opened 2018-02-14
Stop building documentation in Travis

lubomir:
We now use Read The Docs, so there's no point in updating this complex system.

Once this is done, we can redirect github pages to the new documentation site (#150).


hluk:
I think the token used by Travis can be also revoked.

We should replace pdc-client.readthedocs.io (gh-pages) with a simple homepage for pdc-client.

hluk (approved):

PR #152, opened 2018-02-23
Pdc 2570 server configuration

hluk:


pdc_client/__init__.py
```diff
@@ -186,35 +157,21 @@ class to perform requests.
...
             if token is None:
-                token = config.get(CONFIG_TOKEN_KEY_NAME, token)
-        else:
-            url = server
-            if ssl_verify is None:
-                ssl_verify = True
+                token = config.token()
+        except ServerConfigError as e:
+            logger.error(e)
+            sys.exit(1)
```
lubomir: I think this would be better handled in the module that imports `PDCClient`. If someone is using it from their python script, they would not be able to react to this error: it would just kill their script.

pdc_client/config.py
```diff
@@ -0,0 +1,151 @@
...
+            data = json.load(config_file)
+    return data
+
+
+def _default_server_configuration(server):
+    if "://" not in server:
+        message = "Failed to find configuration for server \"{}\"".format(server)
+        raise ServerConfigNotFoundError(message)
+
+    server_config = dict([(CONFIG_URL_KEY_NAME, server)])
```
lubomir: Is `{CONFIG_URL_KEY_NAME: server}` more readable?

pdc_client/config.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class ServerConfig(object):
+    def __init__(self, config):
+        self.config = config
+
+    def get(self, option_name, default_value=None):
+        return self.config.get(option_name, default_value)
+
+    def url(self):
```
lubomir: How about making `url`, `ssl_verify`, `is_development` and `token` properties? Especially `config.ssl_verify()` looks like it should do some verification, but it only returns a value.

pdc_client/config.py
hluk: Yes, indeed. I thought that's not possible in Python - I must be thinking about different language.

pdc_client/config.py
hluk: Sure, looks nicer. Changed.

pdc_client/__init__.py
hluk: That would break the API in some strange way. But ... I guess calling exit here is a bug so I'm fixing it.

hluk:
I finally fixed the Travis builds. I'm spending too much time handling compatibility issues with Python 2.6. Would be nice to finally drop the deprecated Python version.

I removed Python 3.3 from Travis due to some [build errors](https://travis-ci.org/product-definition-center/pdc-client/jobs/345279335).

lubomir (approved):
Dropping Python 2.6 support would be great (and Py 2 altogether even greater), but unfortunately we still have users on RHEL 6.

PR #156, opened 2018-03-02
Sync changes from Fedora dist-git

lubomir:


hluk (approved):
Nice. I like that you kept used same commits with same authors.

PR #157, opened 2018-03-06
Fixes and tests for MockAPI

hluk:


hluk:
MockAPI is used by pdc-updater for tests. It actually doesn't support all features of PDCClient and look bit fragile.

lubomir (approved):

PR #158, opened 2018-03-29
Run tests in spec file

hluk:
---
We had to disable this for some release but I no longer remember why (failing builds, missing dependencies). Anyway, I tested scratch builds with brew (eng-rhel-6) and koji (f27).

### DIFF
--- a/pdc-client.spec
+++ b/pdc-client.spec
@@ -165,6 +165,15 @@ for executable in "%{buildroot}%{_bindir}"/*; do
     "$executable" --help
 done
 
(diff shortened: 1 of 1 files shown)


lubomir (commented):
If both py2 and py3 subpackages are built, should't the tests run for each of them?

hluk:
> If both py2 and py3 subpackages are built, should't the tests run for each of them?

Fixed and tested scratch builds.

lubomir (approved):

